PR #9, opened 2019-11-27
Added IDL and a processing model

yoavweiss:
This adds an IDL for the JS API as well as a processing model


index.bs
```diff
@@ -357,10 +355,46 @@ execute the following steps:
...
+  [SecureContext] Promise&lt;NavigatorUAData&gt; getUserAgent();
+};
+Navigator includes NavigatorUA;
+
+</pre>
 
-ISSUE: Yoav should write all this.
+Processing model {#processing}
+--------------
+<dfn method for="NavigatorUA"><code>getUserAgent()</code></dfn> method *must*
```
mikewest: Nit: Elsewhere in the doc we're using caps (e.g. "MUST").

index.bs
```diff
@@ -357,10 +355,46 @@ execute the following steps:
...
+    agent's [=user agent/platform architecture=].
+5. Set |UAData|'s {{NavigatorUAData/model}} to the value of the user agent's [=user agent/model=].
+6. Set |UAData|'s {{NavigatorUAData/mobile}} to the value of the user agent's
+    [=user agent/mobileness=].
+7. Let |version| be the value of the user agent's [=user agent/major version=].
+8. Let |settings| be the [=context object=]’s [=relevant settings object=].
+9. If |settings|'s [=environment settings object/client-hints set=] [=list/contains=] `UA`, then
+    set |version| to be the value of the user agent's [=user agent/full version=].
+10. Set |UAData|'s {{NavigatorUAData/version}} to the value of |version|.
+11. Return |UAData|.
```
mikewest: We need to return a promise that resolves to |UAData| "in parallel".

index.bs
```diff
@@ -357,10 +355,46 @@ execute the following steps:
...
+3. Set |UAData|'s {{NavigatorUAData/platform}} to the value of the user agent's
+    [=user agent/platform brand and version=].
+4. Set |UAData|'s {{NavigatorUAData/architecture}} to the value of the user
+    agent's [=user agent/platform architecture=].
+5. Set |UAData|'s {{NavigatorUAData/model}} to the value of the user agent's [=user agent/model=].
+6. Set |UAData|'s {{NavigatorUAData/mobile}} to the value of the user agent's
+    [=user agent/mobileness=].
+7. Let |version| be the value of the user agent's [=user agent/major version=].
+8. Let |settings| be the [=context object=]’s [=relevant settings object=].
+9. If |settings|'s [=environment settings object/client-hints set=] [=list/contains=] `UA`, then
```
mikewest: Why do you check the client-hints set list here, and nowhere else? Shouldn't we check all of them or none of them?

index.bs
yoavweiss: ack

index.bs
yoavweiss: There's no need to check the client-hints set for the JS variables, is there?
Is there a way to get the only the major version in JS? (to avoid touching too many bits)

index.bs
yoavweiss: Yup!

index.bs
mikewest: I'm wondering why we're checking the client-hints set at all. If we're exposing platform and model and etc. without checking, why are we not exposing the full version here? Is your goal to ensure that developers can use less of their privacy budget by only asking for the major version?

yoavweiss: Removed the check and added an issue about adding an extra method for major version

yoavweiss: @mikewest - and answering your question that I just now saw after replying (Thanks GitHub!), we don't really need this check, and I got myself confused with the header behavior. At the same time, might be a good idea to allow checking just the major version "for free". Opened an issue

index.bs
```diff
@@ -357,10 +355,46 @@ execute the following steps:
...
+4. Set |UAData|'s {{NavigatorUAData/architecture}} to the value of the user
+    agent's [=user agent/platform architecture=].
+5. Set |UAData|'s {{NavigatorUAData/model}} to the value of the user agent's [=user agent/model=].
+6. Set |UAData|'s {{NavigatorUAData/mobile}} to the value of the user agent's
+    [=user agent/mobileness=].
+7. Let |version| be the value of the user agent's [=user agent/full version=].
+8. Set |UAData|'s {{NavigatorUAData/version}} to the value of |version|.
+9. Let |promise| be [=a new promise=].
+10. Resolve |promise| with |UAData| [=in parallel=].
+11. Return |promise|.
```
mikewest: I think this could benefit from a little restructuring, like:

```
1.  Let |p| be a [=a new promise=].

2.  Run the following steps [=in parallel=]:

    1.  Let |UAData| be a new {{NavigatorUAData}} object whose values are initialized as follows:

        :   {{NavigatorUAData/brand}}
        ::  The user agent's [=user agent/brand=].
        :   {{NavigatorUAData/platform}}
        ::  The user agent's [=user agent/platform brand and version=].
        ...
        :   {{NavigatorUAData/version}}
        ::  The user agent's [=user agent/full version=].

    2.  [=Resolve=] |promise| with |UAData|.

3.  Return |p|.
```

yoavweiss: Done!

PR #10, opened 2019-11-27
Cleanup broken refs

yoavweiss:


yoavweiss:
It is "bikeshed problems". Bikeshed passes, but botches the refs: https://wicg.github.io/ua-client-hints/#fetch-integration

yoavweiss:
@tabatkins - any advice here? It seems like there are two issues with these refs:
1) There are issues with refs that include `code` in them
2) Lowercasing of the first letter in the definition but not the call seems to throw Bikeshed off

mikewest:
It looks like Bikeshed processes the code bits prior to the `[$...$]` shorthand, and then iterates through text nodes. So the shorthand is split, and `[$Set the` doesn't match the regex.

Reworking this as follows works:

1.  Replace ```To <dfn abstract-op>set the `Sec-CH-Arch` header for a request</dfn>, given a [=request=] (|r|)``` with ```To <dfn abstract-op local-lt="set-arch">set the `Sec-CH-Arch` header for a request</dfn>, given a [=request=] (|r|),``` (just add the `local-lt="..."` bit).

2.  Replace ```[$Set the `Sec-CH-Arch` header for a request$], given |r|.``` with ```<a abstract-op lt="set-arch">Set the `Sec-CH-Arch` header for a request</a>, given |r|.```.

WDYT?

tabatkins:
Argh, yeah, unfortunately shorthands currently depend on being present fully in a single text node; if they're split by elements (or other shorthands that happen to be processed earlier in the order...) they break. This has been a long-standing issue that I plan to fix by rewriting the parser in general. :(

yoavweiss:
> It looks like Bikeshed processes the code bits prior to the `[$...$]` shorthand, and then iterates through text nodes. So the shorthand is split, and `[$Set the` doesn't match the regex.
> 
> Reworking this as follows works:
> 
> 1. Replace `` To <dfn abstract-op>set the `Sec-CH-Arch` header for a request</dfn>, given a [=request=] (|r|) `` with `` To <dfn abstract-op local-lt="set-arch">set the `Sec-CH-Arch` header for a request</dfn>, given a [=request=] (|r|), `` (just add the `local-lt="..."` bit).
> 2. Replace `` [$Set the `Sec-CH-Arch` header for a request$], given |r|. `` with `` <a abstract-op lt="set-arch">Set the `Sec-CH-Arch` header for a request</a>, given |r|. ``.
> 
> WDYT?

Yup! Done

PR #12, opened 2019-11-27
Remove unneeded files

yoavweiss:
Since we've got the deployment process in place, we no longer need `index.html` on the master branch, as it's a source of confusion and merge conflicts.

Also, since we're not pursuing the IETF draft, I thought we could delete it as well.

Let me know what you think regarding the latter.


yoavweiss:
> You can probably remove `Makefile` as well.

Removed! :)

PR #31, opened 2020-01-16
fix a missing close paren

koba04:
This is a tiny fix for a typo.


yoavweiss:
Marked as non substantive for IPR from ash-nazg.

PR #33, opened 2020-01-16
Update README

yoavweiss:
This PR updates the README to be in line with the current proposal

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ browser on iOS) are interesting cases in which the underlying browser engine on 
 doesn't match the engine that the relevant browser built themselves. What should we do in these
 cases?
 
-We have a few options for the string:
+There are a few options for the string:
 
 1.  `"Chrome 73"`, which has the least entropy, but also sets poor expectations.
 2.  `"CriOS 73"` (or `"Chrome on iOS 73"`, or similar) which is basically what's sent today, and categorizes the browser as distinct.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,30 +104,30 @@ accomplish this as follows:
         Note: See the GREASE-like discussion below for how we could anticipate the inevitable lies
         which user agents might want to tell in this field.
 
-    2.  The `Sec-CH-UA-Platform` header field represents the platform's brand and major version. For example:
+    2.  The `Sec-CH-Platform` header field represents the platform's brand and major version. For example:
```
mikewest: Is this a change we're making? I kinda liked that the `UA-` prefix grouped all these together (and some of them are probably too generically named without it).

yoavweiss: I thought so...
As I see it, we have two options:
Either we align all related FP and Accept-CH strings to say "ua-" as well, or we remove it here. But 3 prefixes is starting to be a lot...

README.md
```diff
@@ -247,34 +242,27 @@ browser on iOS) are interesting cases in which the underlying browser engine on
 doesn't match the engine that the relevant browser built themselves. What should we do in these
 cases?
 
-I think we have a few options for the string:
+We have a few options for the string:
```
amtunlimited: Maybe "There are a few..."? Not sure if it really matters all that much.

yoavweiss: Sure

PR #38, opened 2020-01-19
Add "UA-" prefix

yoavweiss:
This PR is aligning the spec with the explainer and implementation, by adding the "UA-" prefix to the hints.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -158,7 +158,7 @@ the underlying platform in the next request:
 
 ``` http
   Sec-CH-UA: "Examplary Browser 73.3R8.2H.1"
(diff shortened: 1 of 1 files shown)


amtunlimited:
LGTM

The comment I made on the other PR reminded me, should we put the "client hint token" associated with each new hint in here somewhere?

yoavweiss:
> The comment I made on the other PR reminded me, should we put the "client hint token" associated with each new hint in here somewhere?

The processing model in [find the client hint value](https://wicg.github.io/client-hints-infrastructure/#find-client-hint-value-section) connects the token to the right headers and values.

Issue #22, opened 2020-01-14
UAs SHOULD have more than a single brand value (Was: Stronger encouragement on set of UA strings)


domenic:
https://github.com/WICG/ua-client-hints#should-the-ua-string-be-a-set indicates an idea of there being a set of user agents.

Although it is unclear why this would be a good idea (#21), if we assume it's desirable, I worry that the current design does not do enough to encourage web developers to treat the UA list as a set. Unless early browser implementations follow this optional part of the spec, I imagine we'll see a lot of parsing code for Sec-CH-UA which assumes there's only a single value. (See also https://github.com/WICG/ua-client-hints/issues/7#issuecomment-574263079.) This will effectively prevent any later browser from ever implementing the optional steps in the spec, because doing so would break server assumptions.

Additionally, the JS API always returns a single string, not a set.

If the intention is to encourage treating the UA as a set of possible UAs, then I think you need to make changes such as:

- Changing Sec-CH-UA to a structured header list of strings, instead of a structured header string.
- Potentially mandating, or at least strongly suggesting, that browsers always include >= 2 items in the list. (Especially early implementations.)
- Change the JS API so that `getUserAgent()` returns a promise for an array of `NavigatorUAData`s, or possibly change `NavigatorUAData`'s `platform` and `version` members to be arrays instead.

yoavweiss:
I believe this is closed by #46. #49 adds various explanations that are also related.

The only thing missing is a SHOULD for including more than a single item.

PR #46, opened 2020-01-23
Turn `UA` into a list

### LINKED ISSUE #21: Set of UA strings needs more explanation
https://github.com/WICG/ua-client-hints#should-the-ua-string-be-a-set and https://wicg.github.io/ua-client-hints/#abstract-opdef-set-the-sec-ch-ua-header-for-a-request step 5 indicate that instead of a single "Brand Version" string, some clients may sometimes send "Brand Version, OtherBrand OtherVersion".

The explainer section doesn't do a good job explaining _why_ any user agent would possibly do this, or why it would work.

Reading between the lines, the idea might be that this is an escape valve so that e.g. Edge 79 could lie and say that it's `Edge 79, Chrome 79` or `Chrome 79, Edge 79`? And then Edge would have to hope that Chrome does the same thing, cooperatively, otherwise sites would just always assume those two strings mean "Edge"? So this is kind of a way of allowing browsers to advertise that they're part of a (rendering-engine-based) equivalence class?

Is the intention that Edge would do this every time it sent `Sec-CH-UA`, or just sometimes? (GREASE seems to be "sometimes".)

Or is the README's more-random examples actually what is intended? Will there be a list of totally-fake browser names that people start using, like the literal example `NotBrowser`? Will this list of fake names be shared (standardized?), or will each browser vendor make up their own?

yoavweiss:
This PR turns the UA into a list of parameterized items (each can contain its own version), and does the same for platform. The goal here is to make parsing on the server easier (as we'd be using Structured Headers) as well as on the client (the attribute returns a FrozenArray of dictionary objects with the information).

Closes #21 and #22

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ challenges in mind with the proposal that follows.
 
 ## A Proposal
 
-By default, servers should receive only the user agent's brand and major version number. Servers
+By default, servers should receive only the user agent's brand and significant version number. Servers
 can opt-into receiving information about minor versions, the underlying operating system's major
 version, and details about the underlying architecture and device. The user agent can make
 reasonable decisions about how to respond to sites' requests for more granular data. We might
@@ -95,10 +95,10 @@ accomplish this as follows:
 
 3.  Browsers should introduce several new Client Hint header fields:
 
-    1.  The `Sec-UA` header field represents the user agent's brand and major version. For example:
+    1.  The `Sec-UA` header field represents the user agent's brand and  version. For example:
 
         ```http
-        Sec-CH-UA: "Chrome 73"
+        Sec-CH-UA: "Chrome; 73"
         ```
 
         Note: See the GREASE-like discussion below for how we could anticipate the inevitable lies
@@ -107,7 +107,7 @@ accomplish this as follows:
     2.  The `Sec-CH-UA-Platform` header field represents the platform's brand and major version. For example:
 
         ```http
-        Sec-CH-UA-Platform: "Win 10"
+        Sec-CH-UA-Platform: "Win; 10"
         ```
 
     3.  The `Sec-CH-UA-Arch` header field represents the underlying architecture's instruction set and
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -62,7 +62,7 @@ challenges in mind with the proposal that follows.
 
 ## A Proposal
 
-By default, servers should receive only the user agent's brand and major version number. Servers
+By default, servers should receive only the user agent's brand and significant version number. Servers
```
amtunlimited: I like the "significant" wording, fitting into the "rolling update" system most modern browsers take.

README.md
```diff
@@ -95,10 +95,10 @@ accomplish this as follows:
 
 3.  Browsers should introduce several new Client Hint header fields:
 
-    1.  The `Sec-UA` header field represents the user agent's brand and major version. For example:
+    1.  The `Sec-UA` header field represents the user agent's brand and  version. For example:
```
amtunlimited: Let's put "significant" here too.

README.md
```diff
@@ -107,7 +107,7 @@ accomplish this as follows:
     2.  The `Sec-CH-UA-Platform` header field represents the platform's brand and major version. For example:
 
         ```http
-        Sec-CH-UA-Platform: "Win 10"
+        Sec-CH-UA-Platform: "Win; v=10"
```
amtunlimited: Do any platforms reveal more-than-significant/non-integer versions to the browser?

README.md
```diff
@@ -157,7 +162,7 @@ accomplish this as follows:
     approach.
 
 User agents will attach the `Sec-CH-UA` header to every secure outgoing request by default, with a value
-that includes only the major version (e.g. "`Chrome 69`"). Servers can opt-into receiving more
+that includes only the significant version (e.g. "`Chrome; 69`"). Servers can opt-into receiving more
```
amtunlimited: Should this be `Chrome; v=69`?

README.md
```diff
@@ -244,14 +256,11 @@ cases?
 
 There are a few options for the string:
 
-1.  `"Chrome 73"`, which has the least entropy, but also sets poor expectations.
-2.  `"CriOS 73"` (or `"Chrome on iOS 73"`, or similar) which is basically what's sent today, and categorizes the browser as distinct.
-3.  `"CriOS 73", "Safari 12"`, which is interesting.
-4.  `"Chrome 73", "Safari 12"`, which is more interesting.
-
+1.  `Chrome; 73`, which has the least entropy, but also sets poor expectations.
```
amtunlimited: Version dict?

index.bs
```diff
@@ -409,13 +420,13 @@ could encourage standardized processing of the `UA` string, by randomly
 including additional, intentionally incorrect, comma-separated entries with
 arbitrary ordering. That would reduce the chance that we ossify on a few
 required strings.  For example, Chrome 73's `Sec-CH-UA` header might be
-`"Chrome 73", "NotBrowser 12"`, `"BrowsingIsFun Version 12b", "Chrome 73"`, or
+`Chrome; 73, NotBrowser; 12`, `BrowsingIsFun; 12b, Chrome; 73`, or
```
amtunlimited: Version dict

README.md
yoavweiss: yup

README.md
yoavweiss: yup

index.bs

README.md
yoavweiss: 👍

yoavweiss:
Thanks for reviewing! :)

PR #48, opened 2020-01-26
Move Promises to guard only the high-entropy bits

### LINKED ISSUE #39: Single promise for getUserAgent?
The proposal suggests this API:
```
interface mixin NavigatorUA {
  [SecureContext] Promise<NavigatorUAData> getUserAgent();
};
```

But there's discussion of potentially requiring a user-prompt for some of the more privacy-sensitive bits. Since the privacy model may be very different for, say, the "brand" and the minor version number, don't we need some ability to request some of the information but not others. I.e. perhaps each property should return it's own promise? That also makes it easier to reason about which data is being used by which sites (i.e. measure the compat risk of breaking/freezing just one field).

yoavweiss:
Closes #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ accomplish this as follows:
 
     ```idl
     interface mixin NavigatorUA {
-      [SecureContext] Promise<NavigatorUAData> getUserAgent();
+      [SecureContext] NavigatorUAData getUserAgent();
     };
 
     Navigator includes NavigatorUA;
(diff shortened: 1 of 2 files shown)


yoavweiss:
@foolip - Could you review this and let me know if the IDL is somehow horribly wrong?

yoavweiss:
@domenic - landing, but I'd still appreciate a review of the IDL

PR #49, opened 2020-01-27
Improved examples for GREASE

### LINKED ISSUE #4: What is a user agent's brand?
https://github.com/WICG/ua-client-hints#a-proposal says:
> By default, servers should receive only the user agent's brand and major version number.

This would be delivered in the `Sec-CH-UA` header as "Chrome 74" in the example given.

Question: What should other Chromium-based browsers send? What everyone does with the UA string is to add something to the end, like "Edge/14.14393" or "OPR/43.0.2442.991".

If the `Sec-CH-UA` header is used like the `User-Agent` header, then a new Chromium-based based browser will have to send the same header as Chrome to get the content that will actually work.

A few problems here:
- How to work around bugs in a popular Chromium-based browser?
- How to collect statistics on which concrete browser (not engine) are used?

Probably the answer is the same mechanism as would be used to get other bits of information in the UA string that have been frozen, like the device model. Are there client hints for "concrete browser"? If there is, and it's practical to use it to work around bugs, will it not be locked in by web compat also? Round and round we go :)

yoavweiss:
Closes #4 

This PR outlines what the UA set may look like and the reasoning behind its potential different forms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,10 +240,19 @@ wrongfully categorizing browsers in the wrong buckets.
 Complementary to that, user agents might encourage standardized processing of
 the `UA` string by randomly including additional, intentionally incorrect,
 comma-separated entries with arbitrary ordering (similar conceptually to TLS's
-[GREASE][4]). Chrome 73's `UA` value might then be `"Chrome"; v="73"`,
-`"NotBrowser"; v="12"`, or `"BrowsingIsFun"; v="12b", "Chrome"; v="73"`.
+[GREASE][4]).
 
-We'd reflect this value in the `NavigatorUAData.brand` attribute.
+Let's examine a few examples:
+* In order to avoid sites from barring unknown browsers from their allow lists, Chrome 73 could send a UA set that includes an non-existent browser, and which varies once in a while.
+  - `"Chrome"; v="73", "NotBrowser"; v="12"`
+* In order to enable equivalence classes based on Chromium versions, Chrome could add the rendering engine and its version to that.
+  - `"Chrome"; v="73", "NotBrowser"; v="12", "Chromium"; v="73"`
+* In order to encourage sites to rely on equivalence classes based on Chromium versions rather than exact UA sniffing, Chrome might remove itself from the set entirely.
+  - `"Chromium"; v="73", "NotBrowser"; v="12"`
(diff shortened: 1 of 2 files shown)


twirl:
I personally don't think that sending varying UA is viable solution.
Our goal is to remove vendor lock, i.e. the situation, when code doesn't work in specific browser for no reason.
To achieve this goal we must NOT send any data regarding specific browser brand.
There are several traits:
  * layout engine (WebKit/Blink/Quantum) and its version
  * JS engine (V8/SpiderMonkey) and its version (or maybe even «ES2017-compatible»-like)
  * build base (i.e. Chromium or Firefox) in case of **standard open source platforms** available for customization

In my opinion, brand names (Chrome, Edge, Opera, whatever) could be used for analytics only, and detecting them in code must be deprecated.

Also, I would suggest simple practice: unknown browsers may simply say they are compatible with specific version of standard engine. Even if they're not actually compatible, it's still better than nothing.

yoavweiss:
> Our goal is to remove vendor lock, i.e. the situation, when code doesn't work in specific browser for no reason.

That's one goal, yes. Another goal is that developers will be able to measure market share in different browsers in hopes it would convince them to test in those browsers. 

> * layout engine (WebKit/Blink/Quantum) and its version
> * JS engine (V8/SpiderMonkey) and its version (or maybe even «ES2017-compatible»-like)
> * build base (i.e. Chromium or Firefox) in case of **standard open source platforms** available for customization

Different browsers that are based on common open source rendering engines are not just skins over those engines. They apply patches, turn flags on and off, etc. Features that work in a certain Chrome version may not work in one of the Chromium-based browsers' equivalent version for product reasons or due to bugs in the patches applied. The inverse may also be true (e.g. a Chromium bug that's patched downstream).

On top of that, exposing explicit build numbers would expose a lot more entropy than desired by default.



> In my opinion, brand names (Chrome, Edge, Opera, whatever) could be used for analytics only, and detecting them in code must be deprecated.

Analytics can be an important use case. Another is bug reporting or login notifications. ("You're logged in from browser X in country Y. Is that really you?")

twirl:
> Analytics can be an important use case. Another is bug reporting or login notifications. ("You're logged in from browser X in country Y. Is that really you?")

That's a valid point, but right here in this proposal it is violated. How this should work, if browser sends fake data, or just ‘NotBrowser’?

PR #65, opened 2020-02-13
Use cases

yoavweiss:
In order to help resolve a few discussions on various issues, I whipped up a list of current use cases for the User-Agent string. I believe we're aiming to preserve most of them, while discouraging a few user-hostile ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,24 +203,25 @@ that the extent of a site's usage can be monitored and evaluated.
 
 # Use-cases
 
-This section attempts to document the current uses for the User-Agent string,
+This section attempts to document the current uses for the `User-Agent` string,
 and how similar functionality could be enabled using UA-CH.
 
 ## Differential serving based on browser features
-This use-case enables services like polyfill.io to serve custom-tailored
-polyfills to their users, without bloating up the experience of users of modern
-browsers.  Similarly, when serving Javascript to users, one can avoid
+This use-case enables services like [polyfill.io](https://polyfill.io) to serve
+custom-tailored polyfills to their users, without bloating up the experience of
+modern browser users.
+Similarly, when serving Javascript to users, one can avoid
 transpilation (which can result in bloat and inefficient code) for browsers
-that support the latest ES features that were used.  Finally, when serving
-images, some browsers don't update their `Accept` request headers, while in
-other cases (*cough* WebP *cough*) the MIME type is not descriptive enough to
-distinguish between different variants of the same format. In those cases,
-knowing the browser and its version can be critical to serving the right image
-variant. 
+that support the latest ES features that were used.
+Finally, when serving images, some browsers don't update their `Accept` request
+headers, while in other cases (*cough* WebP *cough*) the MIME type is not
+descriptive enough to distinguish between different variants of the same
+format. In those cases, knowing the browser and its version can be critical to
+serving the right image variant. 
 
-For that use case to work, the server needs to be aware of the server and its
-meaningful version, in order to map that to a list of available features, and
-know which polyfill or code variant to serve.
+For that use case to work, the server needs to be aware of the browser and its
+meaningful version, and map that to a list of available features. That enables
+it to know which polyfill or code variant to serve.
 
 Services that wish to do that using UA-CH will need to inspect the `Sec-CH-UA`
 header, that is sent by default on every request, and modify their reponse
@@ -228,13 +229,13 @@ based on that.
 
 ## Marketshare Analytics
 A browser's market share can be extremely important. Having visibility into a
-browser's usage can encourage developers to test in that particular browsers,
-ensuring fewer compatibility issues. On top of that, a browser's market share
-can have a direct impact on the browser vendors' business goals, ensuring
-future development of the browser.
+browser's usage can encourage developers to test in that particular browser,
+ensuring fewer compatibility issues for its users. On top of that, a browser's
+market share can have a direct impact on the browser vendors' business goals,
+ensuring future development of the browser.
 
-For that use case to work, the server needs to be aware of the server and its
-meaningful version, in order to be able to register them and find their
+For marketshare analytics to work, the server needs to be aware of the server
+and its meaningful version, in order to be able to register them and find their
 relative market shares.
 
 Sites that wish to provide market share analytics using UA-CH will need to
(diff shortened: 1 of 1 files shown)


yoavweiss:
@jyasskin - Can you take a look?

README.md
```diff
@@ -201,6 +201,219 @@ hints and first-parties can use `Feature-Policy` to decide which third-parties t
...
+descriptive enough to distinguish between different variants of the same
+format. In those cases, knowing the browser and its version can be critical to
+serving the right image variant. 
+
+For that use case to work, the server needs to be aware of the browser and its
+meaningful version, and map that to a list of available features. That enables
+it to know which polyfill or code variant to serve.
+
+Services that wish to do that using UA-CH will need to inspect the `Sec-CH-UA`
+header, that is sent by default on every request, and modify their reponse
```
jyasskin: This assumes that `Sec-CH-UA` accurately reports the browser without any extra opt-in, which some other issues have up in the air. I think this section should list the possible designs, and say how each does or doesn't make the use case possible.

> * If `Sec-CH-UA` can randomly include a different browser's identifier, this use case becomes impossible to support.
> * If `Sec-CH-UA` only includes the browser engine by default, services like `polyfill.io` need to opt into the full UA with `Accept-CH: UA` and inspect the result to determine what response to send.
>    * If the browser supports a TLS extension to opt into client hints in the first round trip, this stays as efficient as today. If there is no such TLS extension, the first response has to include the full polyfill, which is wasteful.
> * If `Sec-CH-UA` includes the actual browser by default, services like `polyfill.io` just need to inspect the `Sec-CH-UA` value to determine what response to send.

README.md
```diff
@@ -201,6 +201,219 @@ hints and first-parties can use `Feature-Policy` to decide which third-parties t
...
+ensuring fewer compatibility issues for its users. On top of that, a browser's
+market share can have a direct impact on the browser vendors' business goals,
+ensuring future development of the browser.
+
+For marketshare analytics to work, the server needs to be aware of the server
+and its meaningful version, in order to be able to register them and find their
+relative market shares.
+
+Sites that wish to provide market share analytics using UA-CH will need to
+inspect the `Sec-CH-UA` header, that is sent by default on every request, and
```
jyasskin: Again, we're not certain the accurate browser should be sent by default on every request. Here one of the options is to send the accurate browser on a fixed fraction of requests, and have the logs analysis divide by that fraction.

README.md
```diff
@@ -201,6 +201,219 @@ hints and first-parties can use `Feature-Policy` to decide which third-parties t
...
+### Low-powered devices
+Some sites serve different content to low powered devices that cannot deal with
+CPU intensive tasks, large video and images, etc.  Such content adaptation
+typically uses the device model information that's integrated in the current
+`User-Agent` string for that purpose, relying on server-side databased to
+convert device models into memory, CPU power, and other categories on which
+they want to split their content.
+
+If the dimension on which the split is made is memory, the Device-Memory Client
+Hint can be used to make that distinction.  Otherwise, with US-CH, sites can
```
jyasskin:
```suggestion
Hint can be used to make that distinction.  Otherwise, with UA-CH, sites can
```

README.md
```diff
@@ -201,6 +201,219 @@ hints and first-parties can use `Feature-Policy` to decide which third-parties t
...
+still retriece the device model by opting in to the `UA-Model` hint.
```
jyasskin:
```suggestion
still retrieve the device model by opting in to the `UA-Model` hint.
```

README.md
```diff
@@ -201,6 +201,219 @@ hints and first-parties can use `Feature-Policy` to decide which third-parties t
...
+logins, and take action in case it's not a login that's done by them or on
+their behalf.
+
+For those notifications to be meaningful, sites need to recognize and
+communicate the commercial brand of the browser to the user. These messages
+often also include the platform and its version in order to make sure the user
+knows which device is in question.
+
+Since such messaging doesn't require any server-side adaptation, it's better
+for this case to use the `getUserAgent` method in order to retrieve the
```
jyasskin: I think @scottlow reported that Edge received bug reports when sites told their users they'd logged in with Chrome. Some sort of "user-understandable browser name" function could help with that ... but that risks having sites make logic decisions based on it, which would recreate all the User-Agent problems. Maybe localizing that string would help? I don't think we have to solve this before shipping UA-CH.

README.md
```diff
@@ -201,6 +201,219 @@ hints and first-parties can use `Feature-Policy` to decide which third-parties t
...
+  required information.
+
+## Conversion modeling
+Some machine learning models use various details from the `User-Agent` string
+in order to estimate various things about users of those user agents.  Similar
```
jyasskin: This seems very close to fingerprinting...

README.md
```diff
@@ -201,6 +201,219 @@ hints and first-parties can use `Feature-Policy` to decide which third-parties t
...
+browser full version and the platform version in order to continue to get
+access to those hints.
+
+## Logs and debugging
+Many services log the `User-Agent` string today and can use it in various ways
+when analyzing past traffic or when trying to debug errors related to their
+service.  Those services will have to use the lower entropy values available
+through `Sec-CH-UA` for logging purposes, or opt-in to receive higher-entropy
+hints. The latter doesn't seem like something services should do just for
+forensic purposes.
```
jyasskin: Although when a user selects "Send Feedback", it might make sense to use the JS functions to collect more information at that point.

README.md
```diff
@@ -201,6 +201,219 @@ hints and first-parties can use `Feature-Policy` to decide which third-parties t
...
+harder*.  Like the case of "spam filtering", it would still be feasible to
+actively collect all the hints about the user as bits of entropy. Unlike the
+above case, this is something that the Privacy Budget is destined to prevent,
+without any alternative mechanisms for persistent user tracking.
+
+## Blocking known bots and crawlers
+Currently, the `User-Agent` string is often used as a brute-force way to block
+known bots and crawlers.  There's a concern that moving "normal" traffic to
+expose less entropy by default will also make it easier for bots to hide in the
+crowd.  While there's some truth to that, that's no reason for making the crowd
```
jyasskin:
```suggestion
crowd.  While there's some truth to that, that's not enough reason for making the crowd
```

README.md
yoavweiss: I don't think the explainer is the place to explore those different options. I prefer to land this current language, discuss the use cases on the various issues and then potentially change the developer advice portion. If that's not something you find reasonable, I can eliminate the developer advice bits for now, and add them once we've come to a consensus on the right approach.

README.md
yoavweiss: It's definitely using a lot of high entropy signals in ways that are similar to fingerprinting. At the same time, I think it's distinct from user tracking or the spam avoidance cases.

README.md
yoavweiss: Yeah, I don't see how such a "give us your real name for realz" API won't get abused...

yoavweiss:
@jyasskin friendly ping! :)

README.md
jyasskin: Having developer advice is definitely better than not. I think the explainer *is* a good place to explore or describe the consequences of different options, but maybe only in the [Considered Alternatives](https://w3ctag.github.io/explainers#considered-alternatives) section. Feel free to merge this and think about a Considered Alternatives section later.

README.md
jyasskin: By translating it, it'd be hard to use? No need to mention it here.

README.md
yoavweiss: SG! Thanks :)

README.md
yoavweiss: Maybe. Let's discuss further at https://github.com/WICG/ua-client-hints/issues/67

README.md
yoavweiss: Good point. Added a comment to that effect.

Issue #56, opened 2020-02-04
Interface for getUserAgent missing Promise


cashpipeplusplus:
In section 3, the interface says:

```js
interface mixin NavigatorUA {
  [SecureContext] NavigatorUAData getUserAgent();
};
```

In section 3.1, the processing model says that `getUserAgent()` returns a `Promise`.

I think the interface needs to be corrected to show a return type of `Promise<NavigatorUAData>` instead of just `NavigatorUAData`.

Thanks!

amtunlimited:
Actually, it should be the other way around. The spec was recently changed so that getUserAgent's components gave promises instead of getUserAgent itself. Either way, pr needed.

petemill:
If `getUserAgent` will no longer return be async, and will have child functions that are async, then does it need to be a function at all?

yoavweiss:
Thanks for reporting! :)
This was obsoleted by #70. Closing

Issue #51, opened 2020-01-27
IDL/API surface review


domenic:
In #48 @yoavweiss asked me to do a check on the IDL. Here's a general issue, that mixes in a bit of API review too.

- `NavigatorUABrandVersionDict` is a bit of a strange name, mainly because it includes `Dict`. This is not web-exposed so it doesn't really matter, but I'd suggest `NavigatorUAPlatform`.

- `required` on `NavigatorUABrandVersionDict`'s `brand` doesn't make sense here, since this is only ever returned, not accepted, by Web IDL APIs. Better to just omit that.

- Why is the array of `NavigatorUABrandVersionDict`s titled `brand`? So you'd access the brand via `navigator.brand[0].brand`? If it's an array, it should have a plural name. And I think a "brand + version" is a "platform" elsewhere, so probably this should be `platforms`?

- `getPlatform()` might be clearer as `getRealPlatform()` or something.

- `navigator.getUserAgent()` should no longer be a function, but instead a readonly attribute, since it's no longer promise-returning or expensive. Which brings up some name collision issues, since the natural name is `navigator.userAgent`... Maybe just `self.userAgent`? Tricky.

- The processing model is totally wrong. `getUserAgent()` is said to return a promise, but it does not in the IDL. The methods and getters are not specced to have appropriate steps for running them; instead they are somehow "initialized" (I guess this means that instead of `navigator.getUserAgent().getPlatform` being a function, it got overwritten, per the spec, with the platform brand and version string?)

yoavweiss:
Thanks for the review!!

Finally getting back to this with https://github.com/WICG/ua-client-hints/pull/70:



> * `NavigatorUABrandVersionDict` is a bit of a strange name, mainly because it includes `Dict`. This is not web-exposed so it doesn't really matter, but I'd suggest `NavigatorUAPlatform`.

Removed the "Dict" part. I'm afraid that "platform" would be overloaded, as we use it here for OS (and have since split the OS brand from version).
Open to renaming tho.





> * `required` on `NavigatorUABrandVersionDict`'s `brand` doesn't make sense here, since this is only ever returned, not accepted, by Web IDL APIs. Better to just omit that.

Omitted.



> * Why is the array of `NavigatorUABrandVersionDict`s titled `brand`? So you'd access the brand via `navigator.brand[0].brand`? If it's an array, it should have a plural name. And I think a "brand + version" is a "platform" elsewhere, so probably this should be `platforms`?

[...]

PR #70, opened 2020-03-02
Single Promise, options object, revamped processing model

yoavweiss:
This addresses the IDL review in #51, feedback from @mikewest on https://chromium-review.googlesource.com/c/chromium/src/+/2019369 as well as #56

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/ua-client-hints/pull/70.html" title="Last updated on Mar 6, 2020, 4:00 PM UTC (e244cac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/70/72cbc30...yoavweiss:e244cac.html" title="Last updated on Mar 6, 2020, 4:00 PM UTC (e244cac)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -339,7 +339,7 @@ dictionary UADataValues {
 interface NavigatorUAData {
   readonly attribute FrozenArray&lt;NavigatorUABrandVersion&gt; UASet;
   readonly attribute boolean mobile;
-  Promise&lt;UADataValues&gt; getHighEntropyValues(optional UADataOptions options);
+  Promise&lt;UADataValues&gt; getHighEntropyValues(optional UADataOptions options = {});
 };
 
 interface mixin NavigatorUA {
@@ -594,4 +594,3 @@ Author/Change controller:
 
 Specification document:
 : this specification ([[#user-agent]]), and Section 5.5.3 of [[RFC7231]]
-
--- a/index.bs
+++ b/index.bs
@@ -322,13 +322,6 @@ dictionary NavigatorUABrandVersion {
   DOMString version;
 };
 
-dictionary UADataOptions {
-  boolean platform = false;
-  boolean platformVersion = false;
-  boolean architecture = false;
-  boolean model = false;
-};
-
 dictionary UADataValues {
   DOMString platform = "";
   DOMString platformVersion = "";
@@ -340,7 +333,7 @@ dictionary UADataValues {
 interface NavigatorUAData {
   readonly attribute FrozenArray&lt;NavigatorUABrandVersion&gt; uaList;
   readonly attribute boolean mobile;
-  Promise&lt;UADataValues&gt; getHighEntropyValues(optional UADataOptions options);
+  Promise&lt;UADataValues&gt; getHighEntropyValues(/* sequence&lt;DOMString&gt */);
 };
 
 interface mixin NavigatorUA {
@@ -380,7 +373,7 @@ On getting, the {{NavigatorUAData/uaList}} attribute MUST run these steps:
 
 On getting, the {{NavigatorUAData/mobile}} attribute must return the user agent's [=user agent/mobileness=].
 
-The <dfn method for="NavigatorUA"><code>getHighEntropyValues()</code></dfn> method, called with {{UADataOptions}} |options|, MUST run these steps:
+The <dfn method for="NavigatorUA"><code>getHighEntropyValues()</code></dfn> method, called with [=sequence<DOMString>=] |hints|, MUST run these steps:
 
 1. Let |p| be a [=a new promise=].
 
@@ -388,13 +381,13 @@ The <dfn method for="NavigatorUA"><code>getHighEntropyValues()</code></dfn> meth
 
     1. Let |uaData| be a new {{UADataValues}}.
 
-    2. If |options|["{{UADataOptions/platform}}"] is true, let |uaData|["{{UADataValues/platform}}"] be the the user agent's [=user agent/platform brand=].
+    2. If |hints| [=contains=] "platform", let |uaData|["{{UADataValues/platform}}"] be the the user agent's [=user agent/platform brand=].
 
-    3. If |options|["{{UADataOptions/platformVersion}}"] is true, let |uaData|["{{UADataValues/platformVersion}}"] be the the user agent's [=user agent/platform version=].
+    3. If |hints| [=contains=] "platformVersion", let |uaData|["{{UADataValues/platformVersion}}"] be the the user agent's [=user agent/platform version=].
 
-    4. If |options|["{{UADataOptions/architecture}}"] is true, let |uaData|["{{UADataValues/architecture}}"] be the the user agent's [=user agent/platform architecture=].
+    4. If |hints| [=contains=] "architecture", let |uaData|["{{UADataValues/architecture}}"] be the the user agent's [=user agent/platform architecture=].
 
-    5. If |options|["{{UADataOptions/model}}"] is true, let |uaData|["{{UADataValues/model}}"] be the the user agent's [=user agent/model=].
+    5. If |hints| [=contains=] "model", let |uaData|["{{UADataValues/model}}"] be the the user agent's [=user agent/model=].
 
     6. [=Queue a task=] to [=resolve=] |p| with |uaData|.
 
--- a/index.bs
+++ b/index.bs
@@ -348,7 +348,9 @@ Processing model {#processing}
 
 <h3 id="monkeypatch-html-windoworworkerglobalscope"><code>WindowOrWorkerGlobalScope</code></h3>
 
-Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA list</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from the [=/list=] returned from running [=create UA list=].
+Each user agent has an associated <dfn>UA list</dfn>, which is a [=/list=] created by running [=create UA list=].
+
+Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA frozen array</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from [=UA list=].
 
 <h3 id="create-ua-list-section">Create a UA list</h3>
 
@@ -378,7 +380,7 @@ When asked to run the <dfn>create UA list</dfn> algorithm, the user agent MUST r
 
 <h3 id="getters">Getters</h3>
 
-On getting, the {{NavigatorUAData/uaList}} attribute MUST return [=this=]'s [=WindowOrWorkerGlobalScope/UA list=].
+On getting, the {{NavigatorUAData/uaList}} attribute MUST return [=this=]'s [=relevant global object=]'s [=WindowOrWorkerGlobalScope/UA frozen array=].
 
 On getting, the {{NavigatorUAData/mobile}} attribute must return the user agent's [=user agent/mobileness=].
 
--- a/index.bs
+++ b/index.bs
@@ -333,7 +333,7 @@ dictionary UADataValues {
 interface NavigatorUAData {
   readonly attribute FrozenArray&lt;NavigatorUABrandVersion&gt; uaList;
   readonly attribute boolean mobile;
-  Promise&lt;UADataValues&gt; getHighEntropyValues(sequence&lt;DOMString&gt hints);
+  Promise&lt;UADataValues&gt; getHighEntropyValues(sequence&lt;DOMString&gt; hints);
 };
 
 interface mixin NavigatorUA {
--- a/index.bs
+++ b/index.bs
@@ -350,7 +350,7 @@ Processing model {#processing}
 
 Each user agent has an associated <dfn>UA list</dfn>, which is a [=/list=] created by running [=create UA list=].
 
-Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA frozen array</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from [=UA list=].
+Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA frozen array</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from the [=user agent=]'s [=UA list=].
 
 <h3 id="create-ua-list-section">Create a UA list</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -323,10 +323,10 @@ dictionary NavigatorUABrandVersion {
 };
 
 dictionary UADataValues {
-  DOMString platform = "";
-  DOMString platformVersion = "";
-  DOMString architecture = "";
-  DOMString model = "";
+  DOMString platform; 
+  DOMString platformVersion;
+  DOMString architecture;
+  DOMString model;
 };
 
 [Exposed=Window]
@@ -346,8 +346,13 @@ Navigator includes NavigatorUA;
 Processing model {#processing}
 --------------
 
-On getting, the {{NavigatorUAData/uaList}} attribute MUST run these steps:
+<h3 id="monkeypatch-html-windoworworkerglobalscope"><code>WindowOrWorkerGlobalScope</code></h3>
 
+Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA list</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from the [=/list=] returned from running [=create UA list=].
+
+<h3 id="create-ua-list-section">Create a UA list</h3>
+
+When asked to run the <dfn>create UA list</dfn> algorithm, the user agent MUST run the following steps:
 1. Let |list| be a [=/list=].
 
 2. Collect pairs of [=user agent/brand=] and [=user agent/significant version=] which represent the user agent,
@@ -371,7 +376,13 @@ On getting, the {{NavigatorUAData/uaList}} attribute MUST run these steps:
 
 5. Return |list|.
 
+<h3 id="getters">Getters</h3>
+
+On getting, the {{NavigatorUAData/uaList}} attribute MUST return [=this=]'s [=WindowOrWorkerGlobalScope/UA list=].
+
 On getting, the {{NavigatorUAData/mobile}} attribute must return the user agent's [=user agent/mobileness=].
+
+<h3 id="getHighEntropyValues"><code>getHighEntropyValues</code> method</h3>
 
 The <dfn method for="NavigatorUA"><code>getHighEntropyValues()</code></dfn> method, called with [=sequence<DOMString>=] |hints|, MUST run these steps:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -316,57 +316,61 @@ Interface {#interface}
...
+  DOMString platform = "";
+  DOMString platformVersion = "";
+  DOMString architecture = "";
+  DOMString model = "";
+};
+
 [Exposed=Window]
 interface NavigatorUAData {
-  readonly attribute FrozenArray&lt;NavigatorUABrandVersionDict&gt; brand;
+  readonly attribute FrozenArray&lt;NavigatorUABrandVersion&gt; UASet;
```
domenic: Should be `uaSet`; attributes start with lowercase.

Separately this seems pretty confusing. Is "ua" now meant to mean "brand + version"? So "ua" is distinct from "ua data" which contains a set of UAs plus a mobile boolean?

index.bs
```diff
@@ -316,57 +316,61 @@ Interface {#interface}
...
   readonly attribute boolean mobile;
-  Promise&lt;DOMString&gt; getPlatform();
-  Promise&lt;SOMString&gt; getPlatformVersion();
-  Promise&lt;DOMString&gt; getArchitecture();
-  Promise&lt;DOMString&gt; getModel();
+  Promise&lt;UADataValues&gt; getHighEntropyValues(optional UADataOptions options);
```
domenic:
```suggestion
  Promise&lt;UADataValues&gt; getHighEntropyValues(optional UADataOptions options = {});
```

(invalid Web IDL)

index.bs
```diff
@@ -316,57 +316,61 @@ Interface {#interface}
...
+  [SecureContext] readonly attribute NavigatorUAData userAgentData;
 };
 Navigator includes NavigatorUA;
 
 </pre>
 
 Processing model {#processing}
 --------------
-<dfn method for="NavigatorUA"><code>getUserAgent()</code></dfn> method MUST run these steps:
+On getting, the {{NavigatorUAData/UASet}} attribute must return the set of {{NavigatorUABrandVersion}} objects, initialized with the user agent's [=user agent/brand=] and [=user agent/significant version=].
```
domenic: This doesn't describe how to go from https://wicg.github.io/ua-client-hints/#user-agent-brand and https://wicg.github.io/ua-client-hints/#user-agent-significant-version to a set. In particular both this and getHighEntryValues() use the same wording.

index.bs
```diff
@@ -316,57 +316,61 @@ Interface {#interface}
...
 
-1. Let |p| be a [=a new promise=].
+On getting, the {{NavigatorUAData/mobile}} attribute must return the user agent's [=user agent/mobileness=].
 
-2.  Run the following steps [=in parallel=]:
+The <dfn method for="NavigatorUA"><code>getHighEntropyValues()</code></dfn> method, optionally called with {{UADataOptions}} |options|, MUST run these steps:
```
domenic: It's always called with the dictionary. The Web IDL layer will default it to the empty dictionary if one is not provided.

index.bs
```diff
@@ -316,57 +316,61 @@ Interface {#interface}
...
-        :   {{NavigatorUAData/getArchitecture}}
-        ::  The user agent's [=user agent/platform architecture=].
-        :   {{NavigatorUAData/getModel}}
-        ::  The user agent's [=user agent/model=].
-        :   {{NavigatorUAData/mobile}}
-        ::  The user agent's [=user agent/mobileness=].
+2.  Run the following steps [=in parallel=]:
 
-    2.  [=Resolve=] |p| with |UAData|.
+    1. If |options| is the empty object, let |options| be a new {{UADataOptions}}.
```
domenic: This step is not necessary, per the above.

index.bs
```diff
@@ -316,57 +316,61 @@ Interface {#interface}
...
+    2. Let |UAData| be a new {{UADataValues}}.
+    3. If |options.platform| is true, let |UAData.platform| be the the user agent's [=user agent/platform brand=].
```
domenic: The syntax we use here these days is `|options|["{{UADataValues/platform}}"]`. (Similar for the |UAData|.)

index.bs
```diff
@@ -316,57 +316,61 @@ Interface {#interface}
...
-        ::  The user agent's [=user agent/platform architecture=].
-        :   {{NavigatorUAData/getModel}}
-        ::  The user agent's [=user agent/model=].
-        :   {{NavigatorUAData/mobile}}
-        ::  The user agent's [=user agent/mobileness=].
+2.  Run the following steps [=in parallel=]:
 
-    2.  [=Resolve=] |p| with |UAData|.
+    1. If |options| is the empty object, let |options| be a new {{UADataOptions}}.
+    2. Let |UAData| be a new {{UADataValues}}.
```
domenic: Nit: variable names should be lowercase, so `|uaData|`.

index.bs
```diff
@@ -316,57 +316,61 @@ Interface {#interface}
...
+    3. If |options.platform| is true, let |UAData.platform| be the the user agent's [=user agent/platform brand=].
+    4. If |options.platformVersion| is true, let |UAData.platformVersion| be the the user agent's [=user agent/platform version=].
+    5. If |options.architecture| is true, let |UAData.architecture| be the the user agent's [=user agent/platform architecture=].
+    6. If |options.model| is true, let |UAData.model| be the the user agent's [=user agent/model=].
+    7. [=Resolve=] |p| with |UAData|.
```
domenic: You need to queue a task on a specific task source to resolve this promise.

index.bs
yoavweiss: OK, cool

index.bs
yoavweiss: ok

index.bs
yoavweiss: OK

index.bs
yoavweiss: cool

index.bs
yoavweiss: yeah. Added a more explicit algorithm.

index.bs
yoavweiss: Queued the task. I'm not sure which queue from the [generic task queues](https://html.spec.whatwg.org/multipage/webappapis.html#generic-task-sources) would be the appropriate one. Added an ISSUE to that effect

index.bs
yoavweiss: Renamed to uaSet. Agree that the naming can be clearer. I'll open a separate issue to discuss it.

index.bs
yoavweiss: Bikeshed fails to build for me with the `= {}` bit...

mikewest: Can you file a bug against bikeshed, and run with `/* = {} */` until it's fixed?

/cc @tabatkins

index.bs
```diff
@@ -316,56 +317,90 @@ Interface {#interface}
...
 
 <pre class="idl">
-dictionary NavigatorUABrandVersionDict {
-  required DOMString brand;
+dictionary NavigatorUABrandVersion {
+  DOMString brand;
   DOMString version;
 };
 
+dictionary UADataOptions {
```
mikewest: Nit: I'm fine with this. I'd also be fine with passing an array of names (e.g. `['platform', 'platformVersion', 'architecture']`). The dictionary approach makes sense to me in a world where we expect one or more of these to be non-boolean (e.g. they need some options). If we expect to match the rest of client hints insofar as the name maps directly to a thing without options, the array makes more sense to me.

index.bs
```diff
@@ -316,56 +317,90 @@ Interface {#interface}
 =================
 
 <pre class="idl">
-dictionary NavigatorUABrandVersionDict {
-  required DOMString brand;
+dictionary NavigatorUABrandVersion {
+  DOMString brand;
   DOMString version;
```
mikewest: Is this the full version, or the major version? If the former, the latter should be in the high-entropy group. If the latter, should we expose the former distinctly so developers can make a choice about what data they access above and beyond what we give them by default?

index.bs

index.bs
yoavweiss: OK. I originally went with an options object as it's presumably lower overhead, but this is not likely to be called multiple times, so it probably doesn't matter.

index.bs
yoavweiss: This is the major version. I intend to add a full version API (to the high entropy group) in a followup

index.bs
```diff
@@ -316,56 +317,83 @@ Interface {#interface}
...
 [Exposed=Window]
 interface NavigatorUAData {
-  readonly attribute FrozenArray&lt;NavigatorUABrandVersionDict&gt; brand;
+  readonly attribute FrozenArray&lt;NavigatorUABrandVersion&gt; uaList;
   readonly attribute boolean mobile;
-  Promise&lt;DOMString&gt; getPlatform();
-  Promise&lt;SOMString&gt; getPlatformVersion();
-  Promise&lt;DOMString&gt; getArchitecture();
-  Promise&lt;DOMString&gt; getModel();
+  Promise&lt;UADataValues&gt; getHighEntropyValues(/* sequence&lt;DOMString&gt */);
```
yoavweiss: Bikeshed is giving me " IDL SYNTAX ERROR" if the comment is not there. Am I holding it wrong?

yoavweiss:
@domenic @mikewest PTAL?

index.bs
```diff
@@ -316,56 +317,83 @@ Interface {#interface}
...
 
-3.  Return |p|.
+    4. If |hints| [=contains=] "architecture", let |uaData|["{{UADataValues/architecture}}"] be the the user agent's [=user agent/platform architecture=].
 
-ISSUE: Provide a method to only access the UA's significant version.
+    5. If |hints| [=contains=] "model", let |uaData|["{{UADataValues/model}}"] be the the user agent's [=user agent/model=].
+
+    6. [=Queue a task=] to [=resolve=] |p| with |uaData|.
+
+ISSUE: Add a specific task source this is queued on.
```
yoavweiss: Thoughts on the task queue for this?

index.bs
yoavweiss: Dumb error... Fixed

index.bs
```diff
@@ -316,56 +317,83 @@ Interface {#interface}
...
 <pre class="idl">
-dictionary NavigatorUABrandVersionDict {
-  required DOMString brand;
+dictionary NavigatorUABrandVersion {
+  DOMString brand;
   DOMString version;
 };
 
+dictionary UADataValues {
+  DOMString platform = "";
```
domenic: Since these are only return values they don't need default values.

index.bs
```diff
@@ -316,56 +317,83 @@ Interface {#interface}
...
 [Exposed=Window]
 interface NavigatorUAData {
-  readonly attribute FrozenArray&lt;NavigatorUABrandVersionDict&gt; brand;
+  readonly attribute FrozenArray&lt;NavigatorUABrandVersion&gt; uaList;
   readonly attribute boolean mobile;
-  Promise&lt;DOMString&gt; getPlatform();
-  Promise&lt;SOMString&gt; getPlatformVersion();
-  Promise&lt;DOMString&gt; getArchitecture();
-  Promise&lt;DOMString&gt; getModel();
+  Promise&lt;UADataValues&gt; getHighEntropyValues(sequence&lt;DOMString&gt hints);
```
domenic:
```suggestion
  Promise&lt;UADataValues&gt; getHighEntropyValues(sequence&lt;DOMString&gt; hints);
```

for tidiness

index.bs
```diff
@@ -316,56 +317,83 @@ Interface {#interface}
...
 };
 Navigator includes NavigatorUA;
 
 </pre>
 
 Processing model {#processing}
 --------------
-<dfn method for="NavigatorUA"><code>getUserAgent()</code></dfn> method MUST run these steps:
+
+On getting, the {{NavigatorUAData/uaList}} attribute MUST run these steps:
```
domenic: As written, this will return a new frozen array each time. That's not great. In particular `navigator.userAgentData.uaList !== navigator.userAgentData.uaList`.

The way to do this is to have an associated `<dfn for="NavigatorUAData">UA list</dfn>` and say that it must be initialized following this algorithm, plus at the end, a call to https://heycam.github.io/webidl/#dfn-create-frozen-array. Then, the getting algorithm for `{{NavigatorUAData/uaList}}` returns this's UA list.

You can see an example of this in https://wicg.github.io/origin-policy/#monkeypatch-html-windoworworkerglobalscope. Here you can get away without an explicit initialization point (i.e. no need to monkeypatch navigation or similar to set the UA list) since the algorithm is the same for all NavigatorUAData instances.

Oh, that raises a new point. Do you want to allow the answer to vary between NavigatorUAData instances, or do you need to maintain consistency across some scope? E.g. across the UA, or across Windows that can reach each other synchronously, or...

index.bs
```diff
@@ -316,56 +317,83 @@ Interface {#interface}
...
-        ::  The user agent's [=user agent/platform brand=].
-        :   {{NavigatorUAData/getPlatformVersion}}
-        ::  The user agent's [=user agent/platform version=].
-        :   {{NavigatorUAData/getArchitecture}}
-        ::  The user agent's [=user agent/platform architecture=].
-        :   {{NavigatorUAData/getModel}}
-        ::  The user agent's [=user agent/model=].
-        :   {{NavigatorUAData/mobile}}
-        ::  The user agent's [=user agent/mobileness=].
+    2. If |hints| [=contains=] "platform", let |uaData|["{{UADataValues/platform}}"] be the the user agent's [=user agent/platform brand=].
```
domenic: Nit:

```suggestion
    2. If |hints| [=contains=] "platform", set |uaData|["{{UADataValues/platform}}"] to the the user agent's [=user agent/platform brand=].
```

(same for the others)

index.bs
domenic: What do you plan to implement in Chrome?

There's always the option of giving maximum flexibility by creating an entirely new task source.

The closest counterpart I can think of off the top of my head is createImageBitmap. Unfortunately that doesn't do this right either :(. https://github.com/whatwg/html/issues/5329

index.bs
yoavweiss: OK

index.bs
yoavweiss: Thanks! Did the above, and did patch WindowOrWorkerGlobalScope, to make sure consistency is maintained across them. 

I don't have a strong opinion about consistency across the UA or across Windows - it seems important that the algorithm would end up with the same values for them (especially with headers, to reduce caching variance), but not super important (to me) that objects would `===`.
At the same time, if there's an easy way to make that happen in spec and implementation, I'm happy to do that.

domenic: Oh, I don't think the objects should be ===; that'd be a bit strange. I was more concerned about whether the values should be consistent. I'll re-review from that perspective.

index.bs
```diff
@@ -316,56 +317,94 @@ Interface {#interface}
...
+
+<h3 id="getters">Getters</h3>
+
+On getting, the {{NavigatorUAData/uaList}} attribute MUST return [=this=]'s [=WindowOrWorkerGlobalScope/UA list=].
+
+On getting, the {{NavigatorUAData/mobile}} attribute must return the user agent's [=user agent/mobileness=].
+
+<h3 id="getHighEntropyValues"><code>getHighEntropyValues</code> method</h3>
+
+The <dfn method for="NavigatorUA"><code>getHighEntropyValues()</code></dfn> method, called with [=sequence<DOMString>=] |hints|, MUST run these steps:
```
domenic: Nit:
```suggestion
The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></dfn> method MUST run these steps:
```

index.bs
```diff
@@ -316,56 +317,94 @@ Interface {#interface}
...
 
 </pre>
 
 Processing model {#processing}
 --------------
-<dfn method for="NavigatorUA"><code>getUserAgent()</code></dfn> method MUST run these steps:
+
+<h3 id="monkeypatch-html-windoworworkerglobalscope"><code>WindowOrWorkerGlobalScope</code></h3>
+
+Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA list</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from the [=/list=] returned from running [=create UA list=].
```
domenic: So, if you want "create UA list" to run only once per user agent, then you'll need some kind of novel architecture for that. As-written it will be re-run on each Window creation, with potentially different results.

Maybe "Each user agent has a UA list, which is created by running create UA list." Ideally you'd clarify that this is done once per ... user agent session?? ... until the browser gets updated??? ... but that doesn't seem strictly necessary, as just associating it with the UA will make it somewhat clear it should not get reset (e.g. reset on each WindowOrWorkerGlobalScope creation). 

Then, each WindowOrWorkerGlobalScope can have a UA list (or, UA list frozen array?), which is the result of running create a frozen array on the user agent's UA list.

Then, you can update other parts of the spec (e.g. the headers) to reference the user agent's UA list.

index.bs
```diff
@@ -316,56 +317,94 @@ Interface {#interface}
...
+
+    2.  Randomize the order of the items in |list|.
+
+    Note: See [[#grease]] for more details on why these steps might be appropriate.
+
+5. Return |list|.
+
+<h3 id="getters">Getters</h3>
+
+On getting, the {{NavigatorUAData/uaList}} attribute MUST return [=this=]'s [=WindowOrWorkerGlobalScope/UA list=].
```
domenic: this is a NavigatorUAData, not a WindowOrWorkerGlobalScope. You can either move the UA list to NavigatorUAData, or you can use this's relevant global object's UA list.

yoavweiss: yeah, missed that. Will change to the latter option

index.bs
yoavweiss: done

index.bs
```diff
@@ -316,56 +317,96 @@ Interface {#interface}
...
 
 Processing model {#processing}
 --------------
-<dfn method for="NavigatorUA"><code>getUserAgent()</code></dfn> method MUST run these steps:
+
+<h3 id="monkeypatch-html-windoworworkerglobalscope"><code>WindowOrWorkerGlobalScope</code></h3>
+
+Each user agent has an associated <dfn>UA list</dfn>, which is a [=/list=] created by running [=create UA list=].
+
+Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA frozen array</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from [=UA list=].
```
domenic:
```suggestion
Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA frozen array</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from the [=user agent=]'s [=UA list=].
```

index.bs
```diff
@@ -316,56 +317,96 @@ Interface {#interface}
...
 
 </pre>
 
 Processing model {#processing}
 --------------
-<dfn method for="NavigatorUA"><code>getUserAgent()</code></dfn> method MUST run these steps:
+
+<h3 id="monkeypatch-html-windoworworkerglobalscope"><code>WindowOrWorkerGlobalScope</code></h3>
+
+Each user agent has an associated <dfn>UA list</dfn>, which is a [=/list=] created by running [=create UA list=].
```
domenic:
```suggestion
Each [=user agent=] has an associated <dfn>UA list</dfn>, which is a [=/list=] created by running [=create UA list=].
```

index.bs
```diff
@@ -316,56 +317,96 @@ Interface {#interface}
...
+
+Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrWorkerGlobalScope">UA frozen array</dfn>, which is a <code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is initially the result of [=create a frozen array|creating a frozen array=] from [=UA list=].
+
+<h3 id="create-ua-list-section">Create a UA list</h3>
+
+When asked to run the <dfn>create UA list</dfn> algorithm, the user agent MUST run the following steps:
```
domenic: Nit: the spec is inconsistent about whether it uses MUST or must. As the editor it's your call, but I personally prefer following https://infra.spec.whatwg.org/#conformance and sticking with lowercase.

index.bs
domenic: You could do this throughout the document I guess. Maybe not worth bothering with in this PR.

index.bs
yoavweiss: I'll fix that and the `[=user agent=]` comment everywhere in a followup

PR #74, opened 2020-03-05
Add missing 'CH'

rowan-m:
That's it. Just spotted a typo while reading.


yoavweiss:
Marked as non substantive for IPR from ash-nazg.

yoavweiss:
Thanks for fixing!!
Marked this as non-substantive as its a typo and the IPR bots complained. You should probably join the WICG and link your GH account to your W3C one, to prevent future IPR bots unhappiness.

Issue #45, opened 2020-01-23
Accept-CH for UA to get full version is weird


yoavweiss:
The current mechanism where opting-in to "UA" gives you the same headers only with different values is something I found difficult to explain.
A completely separate opt-in to receive the full version seems more ergonomic, easy to explain and would also solve #11

yoavweiss:
I think a cleaner mechanism would be a separate CH and JS API to expose the full version. See #11 

@mikewest - Thoughts?

yoavweiss:
Fixed with #79

PR #79, opened 2020-03-06
Add a "full version" hint

### LINKED ISSUE #11: API to expose full version (was: An API to expose only the major version)
The JS API in the proposal only includes a single `version` property, which means there's no "free" way to access the major version information without touching any extra bits. There probably should be.

yoavweiss:
Closes #11 and #45

***

### :boom: Error: write EPROTO 140222018242432:error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version:../deps/openssl/openssl/ssl/s23_clnt.c:772:
 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 16, 2020, 9:20 PM UTC)_.

<details>
<summary>More</summary>

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWICG%2Fua-client-hints%2Fpull%2F79%2Fa55cb91.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fyoavweiss%2Fua-client-hints%2Fpull%2F79.html)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ua-client-hints%2379.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -153,14 +153,15 @@ asks for a little more information by sending an `Accept-CH` header (Section 2.2
 [[I-D.ietf-httpbis-client-hints]]) along with the initial response:
 
 ``` http
-  Accept-CH: UA, Platform
+  Accept-CH: UA-Full-Version, UA-Platform
 ```
 
 In response, the user agent includes more detailed version information, as well as information about
 the underlying platform in the next request:
 
 ``` http
-  Sec-CH-UA: "Examplary Browser"; v="73.3R8.2H.1"
+  Sec-CH-UA: "Examplary Browser"; v="73"
+  Sec-CH-UA-Full-Version: "73.3R8.2H.1"
   Sec-CH-UA-Platform: "Windows"
 ```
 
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -324,10 +339,11 @@ dictionary NavigatorUABrandVersion {
 };
 
 dictionary UADataValues {
-  DOMString platform; 
-  DOMString platformVersion;
-  DOMString architecture;
-  DOMString model;
+  DOMString platform = "";
```
yoavweiss: @domenic - brought back empty string initialization to this returned dictionary, as otherwise, the tests fail (as the non-requested values became `undefined` instead of being an empty strring)

domenic: That must be a Chromium implementation issue? Defaults do not work that way per spec; they only apply when used as an incoming argument. You need to remove it from here and add explicit steps to set them to empty string in the algorithm, if that's the desired behavior.

However, that behavior seems wrong (i.e. the tests seem wrong). Any non-requested fields are way better as being actually not present. Why have them be present with a misleading empty string value?

yoavweiss: Fair enough regarding the default values and I can change the test to look for `undefined`.

Should I also file an issue on Chromium's IDL implementation? 
Looking at https://heycam.github.io/webidl/#dfn-dictionary-member-default-value, can you point me to the specific part that indicates this behavior is wrong?

domenic: Even better would be to have the test check for not present, e.g. `!("platform" in returnValue)`.

https://heycam.github.io/webidl/#es-dictionary "An IDL dictionary value V is converted to an ECMAScript Object value by running the following algorithm (where D is the dictionary):" is the algorithm that's supposed to be implemented here. (Your algorithm is returning an IDL dictionary, which gets converted into an ECMAScript object for exposure to JS consumers.) Step 3.1.2 shows how the ES object only gets a property if the dictionary member is present. If it's not present, then the default value is not consulted; instead, no property is defined.

So yeah, filing a Chromium bug would probably be good :).

yoavweiss: Filed https://bugs.chromium.org/p/chromium/issues/detail?id=1059695 and reverted the initialization

yoavweiss:
@mikewest - friendly ping! :)

yoavweiss:
@morlovich - can you review?

morlovich:
I think the IANA section needs update? Everything else looks fine.

yoavweiss:
@morlovich - thanks for reviewing!

PR #81, opened 2020-03-09
UAs should have more than a single UA entry

### LINKED ISSUE #22: UAs SHOULD have more than a single brand value (Was: Stronger encouragement on set of UA strings)

yoavweiss:
Adds a requirement to have more than a single entry in the UA set (as well as tidies up)

Closes #22

***

### :boom: Error: write EPROTO 140222018242432:error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version:../deps/openssl/openssl/ssl/s23_clnt.c:772:
 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 16, 2020, 4:54 PM UTC)_.

<details>
<summary>More</summary>

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWICG%2Fua-client-hints%2Fpull%2F81%2Ffb3f94e.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fyoavweiss%2Fua-client-hints%2Fpull%2F81.html)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ua-client-hints%2381.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -461,31 +461,33 @@ order to reduce the fingerprint the header provides.
 GREASE-like UA Strings {#grease}
 ----------------------
 
-History has shown us that there are real incentives for [=user agents=] to lie
-about their branding in order to thread the needle of sites' sniffing scripts,
-and prevent their users from being blocked by UA-based allow/block lists.
+History has shown us that there are real incentives for [=user agents=] to lie about their branding
+in order to thread the needle of sites' sniffing scripts, and prevent their users from being blocked
+by UA-based allow/block lists.
 
-Reseting expectations may help to prevent abuse of the UA string's brand in the
-short term, but probably won't help in the long run.  The world of network
-protocol introduced the notion of <abbr title="Generate Random Extensions And
-Sustain Extensibility">GREASE</abbr> [[I-D.ietf-tls-grease]]. We could borrow
-from that concept to tackle this problem.
+Reseting expectations may help to prevent abuse of the UA string's brand in the short term, but
+probably won't help in the long run.  The world of network protocol introduced the notion of <abbr
+title="Generate Random Extensions And Sustain Extensibility">GREASE</abbr> [[I-D.ietf-tls-grease]].
+We could borrow from that concept to tackle this problem.
 
-[=User agents=] SHOULD model `UA` as a set, rather than a single entry. They then
-could encourage standardized processing of the `UA` string, by randomly
-including additional, intentionally incorrect, comma-separated entries with
-arbitrary ordering. That would reduce the chance that we ossify on a few
-required strings.  
+[=User agents=] SHOULD model <a http-header>`Sec-CH-UA`</a> and {{uaList}} as a set which contains
+more than a single entry. They then could encourage standardized processing of the `UA` string, by
+randomly including additional, intentionally incorrect, comma-separated entries with arbitrary
+ordering. That would reduce the chance that we ossify on a few required strings.  
 
 Let's examine a few examples:
-* In order to avoid sites from barring unknown browsers from their allow lists, Chrome could send a UA set that includes an non-existent browser, and which varies once in a while.
-  - `"Chrome"; v="73", "NotBrowser"; v="12"`
-* In order to enable equivalence classes based on Chromium versions, Chrome could add the rendering engine and its version to that.
-  - `"Chrome"; v="73", "NotBrowser"; v="12", "Chromium"; v="73"`
-* In order to encourage sites to rely on equivalence classes based on Chromium versions rather than exact UA sniffing, Chrome might remove itself from the set entirely.
-  - `"Chromium"; v="73", "NotBrowser"; v="12"`
-* Browsers based on Chromium may use a similar UA string, but use their own brand as part of the set, enabling sites to count them.
-  - `"Chrome"; v="73", "Awesome Browser"; v="60", "Chromium"; v="73"`
+* In order to avoid sites from barring unknown browsers from their allow lists, Chrome could send a
+    UA set that includes an non-existent browser, and which varies once in a while.
+    - `"Chrome"; v="73", "NotBrowser"; v="12"`
+* In order to enable equivalence classes based on Chromium versions, Chrome could add the rendering
+    engine and its version to that.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -461,31 +461,33 @@ order to reduce the fingerprint the header provides.
...
+History has shown us that there are real incentives for [=user agents=] to lie about their branding
+in order to thread the needle of sites' sniffing scripts, and prevent their users from being blocked
+by UA-based allow/block lists.
+
+Reseting expectations may help to prevent abuse of the UA string's brand in the short term, but
+probably won't help in the long run.  The world of network protocol introduced the notion of <abbr
+title="Generate Random Extensions And Sustain Extensibility">GREASE</abbr> [[I-D.ietf-tls-grease]].
+We could borrow from that concept to tackle this problem.
+
+[=User agents=] SHOULD model <a http-header>`Sec-CH-UA`</a> and {{uaList}} as a set which contains
```
yoavweiss: I expected GH to highlight the parts that materially changed, vs. just newline differences. But it didn't. So I'm doing it instead.

This sentence here is the only one including substantive differences. The rest is just newline re-jigging.

domenic: I think this should instead refer to `[=user agent/UA list=]`, from which both of these things are derived. And "to return the Sec-CH-UA value for a request" should be updated to use UA list, instead of having a separate algorithm.

Finally, the "create UA list" algorithm should change from a MAY to a SHOULD. And then this section should have no normative requirements, but instead talk about the motivation behind the SHOULD in "create UA list".

yoavweiss: That makes total sense!

index.bs
```diff
@@ -266,34 +266,23 @@ are fairly clearly sniffable by "examining the structure of other headers and by
...
 
-3.  Let |ua| be a string whose value is the [=string/concatenation=] of the [=user agent=]'s
-    [=user agent/brand=], the string ";v=", and |version|.
+    1. Let |parameter| be a [=dictionary=], initially empty.
 
-4.  [=list/Append=] |ua| to |value|.
+    2. Set |parameter|["param_name"] to "v".
 
-5.  The [=user agent=] MAY execute the following steps:
+    3. Set |parameter|["param_value"] to {{NavigatorUABrandVersion/version}}.
```
domenic: These steps should not reference the generic concept of NavigatorUABrandVersion's version/brand fields, but instead a specific instance of it. To do this I think "For each `NavigatorUABrandVersion` in UA list" should become "For each _brandVersion_ in UA list", and then you can reference "_brandVersion_'s `version`".

index.bs
```diff
@@ -266,34 +266,23 @@ are fairly clearly sniffable by "examining the structure of other headers and by
...
 
-    1.  [=list/Append=] additional items to |value| containing arbitrary brand and version
-        combinations.
+    2. Let |pair| be a tuple comprised of {{NavigatorUABrandVersion/brand}} and |parameter|.
 
-    2.  Randomize the order of the items in |value|.
+    3. Append |pair| to |list|.
```
domenic: FWIW you could use literal syntax to shorten these three steps into

> Append (_brandVersion_'s `brand`, «[ "`param_name`" → "`v`", "`param_value`" → _brandVersion_'s `version` ]»).

Not sure if that's clearer, but just wanted to offer it as an option.

(I guess that assumes that structured header's "ordered Dictionary" concept is the same as Infra's "ordered map", but that seems safe.)

index.bs
yoavweiss: That makes sense, thanks! :)

index.bs
yoavweiss: That's good to know. I'll keep it as is, as the current version is clearer IMO

index.bs
```diff
@@ -266,34 +266,23 @@ are fairly clearly sniffable by "examining the structure of other headers and by
 availability and semantics of the features introduced or modified between releases of a particular
 browser" [[Janc2014]]).
 
-To <dfn abstract-op local-lt="set-ua">return the `Sec-CH-UA` value for a request</dfn>, given a
-<a href="https://wicg.github.io/client-hints-infrastructure/#environment-settings-object-client-hints-set">client hints set</a> (|set|),
-[=user agents=] MUST:
+To <dfn abstract-op local-lt="set-ua">return the `Sec-CH-UA` value for a request</dfn>, [=user agents=] MUST:
 
-1.  Let |value| be a [=Structured Header=] object whose value is a [=structured header/list=].
+1.  Let |list| be an [=/list=], initially empty.
```
domenic:
```suggestion
1.  Let |list| be a [=/list=], initially empty.
```

Issue #63, opened 2020-02-08
Should browser version really be exposed in client hints?


mcatanzaro:
Maintaining user agent quirks has become easier since WebKit froze its UA version string at 605.1.15. I hope that, if implementing the UA-CH spec, WebKit would choose to continue sending a fake, frozen version number, and not distinguish between ["significant version" and "full version."](https://wicg.github.io/ua-client-hints/#http-ua-hints) I believe WebKit has a strong interest in making it as difficult as possible for websites to use client hints to alter web content. (Currently the Safari major version number is still increasing, which is a shame. Websites and polyfills should rely on feature detection, not version numbers.)

The stated [rationale for including version information](https://wicg.github.io/ua-client-hints/#intro) is:

> Brand and version information (e.g. "Chrome 69") allows websites to work around known bugs in specific releases that aren’t otherwise detectable. For example, implementations of Content Security Policy have varied wildly between vendors, and it’s difficult to know what policy to send in an HTTP response without knowing what browser is responsible for its parsing and execution.

On the one hand, this seems like a strong argument for why websites would need to alter web content based on browser version. But it's easy to see that, if websites were to really alter CSP based on user agent string, the websites would break badly whenever a small browser sends a browser name and version that doesn't match what the website is expecting to see. It also largely contradicts the purpose of GREASEing the headers: the purpose of the GREASE is to discourage websites from altering web content based on the provided values, but the rationale for including version number only makes sense if websites can indeed rely on the provided values. That's a rather fundamental contradiction.

So I'd much rather expose as little information via client hints as possible. Web compatibility problems with CSP or other web platform features should be treated as bugs and fixed; we shouldn't expect websites to attempt to work around these by inspecting client hints. Including version numbers also creates an annoying game of catch-up for UAs trying to maintain quirks lists. It also encourages websites to blacklist old browser versions and display "unsupported browser" warnings or even entirely block access to the website on the basis of the provided version number, which significantly increases the effort required to maintain a quirks list. Not including this information would make it easier to maintain client hint quirks. (Quirks are inevitably going to be required, GREASE or no GREASE, for everything that's not Chromium.)

Although I'd rather not expose any version number information at all, if the spec does continue to expose versions, it should at least be limited to the "significant version," since full version is an obvious privacy risk and doesn't provide concrete benefit to websites.

CC @othermaciej

mikesherov:
[...]

Issue #58, opened 2020-02-05
Sec-CH-UA-Arch exposes info that is frozen or omitted in some UA strings, thus would be a privacy regression


othermaciej:
In Safari, we have frozen or omitted the CPU value found in some legacy UA strings.

For desktop versions of Safari (including macOS and iPadOS), we use `Intel` and for mobile versions (including iOS and watchOS) we use `CPU`, whether or not this accurately reflects the CPU architecture. And we hav decided not to introduce new values even if we introduce new CPU architectures.

Thus, implementing `Sec-CH-UA-Arch` honestly would be a privacy regression for us. We could implement a fake version that only says `Intel` or `CPU`, or perhaps even a version that only says `CPU`, but then what's the point? It's also not clear if a bogus implementation like this would be compliant with the spec.

I think it would be best if this was dropped, but at the very least it should be optional, since some browsers are choosing not to expose CPU information.

mcatanzaro:
I don't understand what benefit exposing CPU architecture could possibly have. Why would a server possibly care what CPU architecture is being used? (Maybe to autodownload architecture-specific binaries to the user's computer? Is the benefit really worth the cost?)

WebKit has previously experienced trouble where Google websites (e.g. Google Maps, Google Calendar, etc.) provide a degraded user experience when the CPU architecture string is not "x86_64", so we have a quirk (not used by Safari; WebKit and Safari have separate UA quirks) to send an x86_64 user agent to such domains. My guess is that this is some badly misguided attempt to check for OpenGL capability, but it doesn't really matter why it happened: what matters is that major websites actually intentionally look at the CPU architecture portion of the user agent string and improperly decide to degrade the user experience if it doesn't match what they want to see.

Any client hint exposed by this spec is inevitably going to require a quirks list so that we can send fake values to websites that decide to use it to alter web content. It's easier for everyone if we limit client hints to the minimum that's actually required.

othermaciej:
I'm. not sure what it's for either. I guess downloads? On platforms that have multiple CPU architectures and no norm of fat binaries or CPU emulation, executable downloads could be different per-CPU. That said, we've frozen it in Safari and will likely never update it. (We'd probably recommend other ports also freezing at "Intel" for desktop platforms (or "Intel"/"x64"/"x86_64" if it needs to be different per OS) and "CPU" for mobile platforms since these demonstrably work fine.

I would be happiest if this was omitted entirely.

mcatanzaro:
[...]

PR #90, opened 2020-03-16
Clarify what the hints are and that some are optional

### LINKED ISSUE #57: Sec-CH-UA-Model client hint exposes info that is not currently exposed in Use-Agent or otherwise
Sec-CH-UA-Model client hint exposes info that is not currently exposed in User-Agent or otherwise, thus increasing available entropy. The use case is not explained in the spec and is non-obvious.

yoavweiss:
Closes #57, #58 and addresses https://github.com/WICG/ua-client-hints/issues/63#issuecomment-583789879, as well as adds a general description of what the hints are, which was missing.

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/ua-client-hints/pull/90.html" title="Last updated on May 8, 2020, 9:48 PM UTC (b631ae2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/90/ff4d5f0...yoavweiss:b631ae2.html" title="Last updated on May 8, 2020, 9:48 PM UTC (b631ae2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -194,9 +194,17 @@ has defined a number of properties for itself:
 [=User agents=] SHOULD keep these strings short and to the point, but servers MUST accept arbitrary
 values for each, as they are all values constructed at the [=user agent=]'s whim.
 
-[=User agents=] MAY return the empty string for [=full version=], [=platform architecture=] or
-[=model=] in cases they consider that exposing that value increase their user's privacy risk beyond
-the user benefit of the use-cases they enable.
+[=User agents=] SHOULD return the empty string or a fictitious value for [=platform architecture=]
+unless the user's platform is one where both the following conditions apply:
+ * Binary download of executables is likely.
+ * Different CPU architectures are likely to require different binary executable resources.
+
+[=User agents=] MUST return the empty string for [=model=] if [=mobileness=] is false or on
+platforms where the model is not typically exposed.
+
+[=User agents=] MAY return the empty string or a fictitious value for [=full version=], [=platform
+architecture=] or [=model=] in cases they consider that exposing that value increase their user's
+privacy risk beyond the user benefit of the use-cases they enable.
 
 The 'Sec-CH-UA-Arch' Header Field {#sec-ch-arch}
 ------------------------------
--- a/index.bs
+++ b/index.bs
@@ -180,7 +180,7 @@ has defined a number of properties for itself:
       which includes distinguishable web-exposed features (for example: "72", "3", or "12.1")
 *   <dfn for="user agent" export>full version</dfn> - The [=user agent=]'s build version (for
       example: "72.0.3245.12", "3.14159", or "297.70E04154A")
-*   <dfn for="user agent" export>platform brand</dfn> - The [=user agent=]'s operating system
+*   <dfn for="user agent" export>platform brand</dfn> - The [=user agent=]'s operating system's
       commercial name. (for example: "Windows", "iOS", or "AmazingOS")
 *   <dfn for="user agent" export>platform version</dfn> - The [=user agent=]'s operating system's
       version. (for example: "NT 6.0", "15", or "17G")
--- a/index.bs
+++ b/index.bs
@@ -197,14 +197,15 @@ values for each, as they are all values constructed at the [=user agent=]'s whim
 [=User agents=] SHOULD return the empty string or a fictitious value for [=platform architecture=]
 unless the user's platform is one where both the following conditions apply:
  * Binary download of executables is likely.
- * Different CPU architectures are likely to require different binary executable resources.
+ * Different CPU architectures are likely to require different binary executable resources, and
+     different binary exeutable resources are likely to be available.
 
-[=User agents=] MUST return the empty string for [=model=] if [=mobileness=] is false or on
-platforms where the model is not typically exposed.
+[=User Agents=] MUST return the empty string for [=model=] if [=mobileness=] is false. [=User
+Agents=] MUST return the empty string for [=model=] even if [=mobileness=] is true, except on
+platforms where the model is typically exposed.
 
 [=User agents=] MAY return the empty string or a fictitious value for [=full version=], [=platform
-architecture=] or [=model=] in cases they consider that exposing that value increase their user's
-privacy risk beyond the user benefit of the use-cases they enable.
+architecture=] or [=model=], for privacy, compatibility, or other reasons.
 
 The 'Sec-CH-UA-Arch' Header Field {#sec-ch-arch}
 ------------------------------
--- a/index.bs
+++ b/index.bs
@@ -170,22 +170,33 @@ User Agent Hints {#http-ua-hints}
 ================
 
 The following sections define a number of HTTP request header fields that expose detail about a
-given [=user agent=], which servers can opt-into receiving via the Client Hints infrastructure defined
-in [[I-D.ietf-httpbis-client-hints]]. The definitions below assume that each [=user agent=] has defined
-a number of properties for itself:
+given [=user agent=], which servers can opt-into receiving via the Client Hints infrastructure
+defined in [[I-D.ietf-httpbis-client-hints]]. The definitions below assume that each [=user agent=]
+has defined a number of properties for itself:
 
-*   <dfn for="user agent" export>brand</dfn> (for example: "cURL", "Edge", "The World's Best Web Browser")
-*   <dfn for="user agent" export>significant version</dfn> (for example: "72", "3", or "28")
-*   <dfn for="user agent" export>full version</dfn> (for example: "72.0.3245.12", "3.14159", or "297.70E04154A")
-*   <dfn for="user agent" export>platform brand</dfn> (for example: "Windows", "iOS", or "AmazingOS")
-*   <dfn for="user agent" export>platform version</dfn> (for example: "NT 6.0", "15", or "17G")
-*   <dfn for="user agent" export>platform architecture</dfn> (for example: "ARM64", or "ia32")
-*   <dfn for="user agent" export>model</dfn> (for example: "", or "Pixel 2 XL")
-*   <dfn for="user agent" export>mobileness</dfn> (for example: ?0 or ?1)
+*   <dfn for="user agent" export>brand</dfn> - The [=user agent=]'s commercial name (for example:
+      "cURL", "Edge", "The World's Best Web Browser")
+*   <dfn for="user agent" export>significant version</dfn> - The [=user agent=]'s marketing version,
+      which includes distinguishable web-exposed features (for example: "72", "3", or "12.1")
+*   <dfn for="user agent" export>full version</dfn> - The [=user agent=]'s build version (for
+      example: "72.0.3245.12", "3.14159", or "297.70E04154A")
+*   <dfn for="user agent" export>platform brand</dfn> - The [=user agent=]'s operating system
+      commercial name. (for example: "Windows", "iOS", or "AmazingOS")
+*   <dfn for="user agent" export>platform version</dfn> - The [=user agent=]'s operating system's
+      version. (for example: "NT 6.0", "15", or "17G")
+*   <dfn for="user agent" export>platform architecture</dfn> - The [=user agent=]'s underlying CPU
+      architecture (for example: "ARM64", or "ia32")
+*   <dfn for="user agent" export>model</dfn> - The [=user agent=]'s device model (for example: "",
+      or "Pixel 2 XL")
+*   <dfn for="user agent" export>mobileness</dfn> - A boolean indicating if the [=user agent=]'s
+      device is a mobile device. (for example: ?0 or ?1)
 
 [=User agents=] SHOULD keep these strings short and to the point, but servers MUST accept arbitrary
 values for each, as they are all values constructed at the [=user agent=]'s whim.
 
+[=User agents=] MAY return the empty string for [=full version=], [=platform architecture=] or
+[=model=] in cases they consider that exposing that value increase their user's privacy risk beyond
+the user benefit of the use-cases they enable.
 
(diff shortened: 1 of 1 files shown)


yoavweiss:
@morlovich - can you review?

index.bs
```diff
@@ -170,22 +170,33 @@ User Agent Hints {#http-ua-hints}
...
+defined in [[I-D.ietf-httpbis-client-hints]]. The definitions below assume that each [=user agent=]
+has defined a number of properties for itself:
+
+*   <dfn for="user agent" export>brand</dfn> - The [=user agent=]'s commercial name (for example:
+      "cURL", "Edge", "The World's Best Web Browser")
+*   <dfn for="user agent" export>significant version</dfn> - The [=user agent=]'s marketing version,
+      which includes distinguishable web-exposed features (for example: "72", "3", or "12.1")
+*   <dfn for="user agent" export>full version</dfn> - The [=user agent=]'s build version (for
+      example: "72.0.3245.12", "3.14159", or "297.70E04154A")
+*   <dfn for="user agent" export>platform brand</dfn> - The [=user agent=]'s operating system
```
morlovich: system's?

morlovich:
Looks good, though I wonder if OS version is a concern for Android? Right now we also present an entire major.minor.point triplet on that in Chrome.

(And hi, Maciej!)

othermaciej:
While this is an improvement, I think it's too weak, and I would not consider it to address my issues. This PR describes the values, and says user agents may leave them blank " in cases they consider that exposing that value increase their user’s privacy risk beyond the user benefit of the use-cases they enable". This fuzzy rule does not seem like it would lead to interop.

For Arch, my proposal should b:
- User Agents MUST leave this blank in most circumstances, but MAY include a CPU value on platforms where binary downloads from the web are common, and CPU architecture is essential to get the right download. (Currently this includes Windows but probably not most other platforms). 
- User Agents MAY use a fictitious CPU architecture if required for compatibility.
- User Agents MUST (or SHOULD) leave this blank or use a fictitious CPU architecture for uncommon CPU architectures.

For Model, my proposal would be:
- Users Agents MUST leave this blank if the Mobile boolean is false.
- User Agents SHOULD leave this blank even if Mobile boolean is true, unless there is a norm of reporting model on the platform.

For full version I don't have a proposal, but I also did not file an issue about it (yet). Safari would likely not report additional info here. I would propose that if UAs don't want to report extra info, they should report the significant version as the full version.

BTW the fact that the syntax for Mobile is an open-ended string makes it hard to define whether mobileness is true or false. It should be a token list of exact values instead. 0/1, true/false, yes/no, pick one of these pairs. How are servers supposed to parse this if it can be an arbitrary string? I guess I should file a separate issue for that.

othermaciej:
Also Cc @mcatanzaro who commented insightfully on the issues cited.

othermaciej:
Additional comment: I think we should drop "full version" unless there is at least one browser that wants to expose it. Safari would not do so. I guess it sounds like Chrome would indeed want it?

mcatanzaro:
I suggest platform brand and platform version should also be optional. WebKit on Linux would probably set platform brand, but not platform version.

yoavweiss:
(Belatedly getting back to this. Apologies for the delay...)

> * User Agents MUST leave this blank in most circumstances, but MAY include a CPU value on platforms where binary downloads from the web are common, and CPU architecture is essential to get the right download. (Currently this includes Windows but probably not most other platforms).

I don't think we agree here. 
While I accept that on iOS side-loading of binaries is extremely rare, that's not really the case for Android, macOS or Linux. While they have "stores" (or distribution repositories), not everything is there, and users often download binaries elsewhere.
While OS folks may want to discourage that, I don't think we want to rule that use-case as illegitimate.

> * User Agents SHOULD leave this blank even if Mobile boolean is true, unless there is a norm of reporting model on the platform.

"SHOULD" with such a huge caveat feels wrong. 

How about the following:
"[=User agents=] MUST return the empty string for [=platform architecture=] in platforms where binary downloads are unlikely, and MUST return the empty string for [=model=] if [=mobile=] is false or on platforms where the model is not typically exposed.
[=User agents=] MAY return the empty string or a fictitious value for [=platform architecture=] or [=model=], and MAY return the [=significant version=] value for [=full version=] in case they consider that exposing the real values increases their users' privacy risk beyond the user benefit of the use-cases they enable."





> BTW the fact that the syntax for Mobile is an open-ended string makes it hard to define whether mobileness is true or false. It should be a token list of exact values instead. 0/1, true/false, yes/no, pick one of these pairs. How are servers supposed to parse this if it can be an arbitrary string? I guess I should file a separate issue for that.

The [mobile value is a boolean](https://wicg.github.io/ua-client-hints/#sec-ch-mobile), at least at the moment. Maybe there was a typo there that was since fixed?

yoavweiss:
> Additional comment: I think we should drop "full version" entirely unless there is at least one browser that wants to expose it. Safari would not do so. I guess it sounds like Chrome would indeed want it?

Chrome would want to expose it in order to match the current capabilities of the User-Agent string. Maybe over time we can find a better alternative to tackle the [use cases](https://github.com/WICG/ua-client-hints#vulnerability-filtering) it enables. If so, we could deprecate and remove it then.

yoavweiss:
> I suggest platform brand and platform version should also be optional. WebKit on Linux would probably set platform brand, but not platform version.

Sure. If the overall suggested text above makes sense, I could also squeeze platform version in as an optional value as well. (It [seems](https://developers.whatismybrowser.com/useragents/parse/4499-webkit-based-browser-linux-webkit) like it's not currently exposed, so making it optional makes sense)

gmeroz:
What’s the reason not to have platform brand and model mandatory?

These fields are extremely useful to detect bots and ad fraud using anomaly detection tools

yoavweiss:
> What’s the reason not to have platform brand and model mandatory? These fields are extremely useful to detect bots and ad fraud using anomaly detection tools

While it's certainly useful for [spam filtering](https://github.com/WICG/ua-client-hints#spam-filtering-and-bot-detection), it's also useful for [persistent user tracking](https://github.com/WICG/ua-client-hints#persistent-user-tracking), which is a use-case we explicitly want to prevent.

So while it makes sense for browsers that currently passively expose that information to continue to expose it through UA-CH, it may not make sense for browsers that don't currently expose it to *start* exposing it.

othermaciej:
> I don't think we agree here.
While I accept that on iOS side-loading of binaries is extremely rare, that's not really the case for Android, macOS or Linux. While they have "stores" (or distribution repositories), not everything is there, and users often download binaries elsewhere.
> While OS folks may want to discourage that, I don't think we want to rule that use-case as illegitimate.

Are there real websites that offer different CPU-specific downloads for macOS, Linux or Android? Or is the purpose here just to do a lockout if the user appears to have the "wrong" CPU? I think this is only a real thing for Windows. Given real examplesI might believe the others. I don't think its right to suggest that needing the exact CPU is the norm and hiding it is the rare exception.

> "[=User agents=] MUST return the empty string for [=platform architecture=] in platforms where binary downloads are unlikely, 

I don't think "where binary downloads are unlikely" is sufficient. It ought to be be left blank on platforms where different binary downloads for different CPUs are unlikely. The expanded condition applies for example on macOS, where even though binary downloads are likely, x86_64 binaries or fat binaries would be the norm, so exposing CPU architecture would not be useful.

>and MUST return the empty string for [=model=] if [=mobile=] is false or on platforms where the model is not typically exposed.

I think it's clearer to make these separate sentences instead of linking with "and", since they are unrelated. I think the is weirdly backwards. The only platform where the model is typically exposed afaik is Android. I'd write it as "User Agents MUST empty string for [=model=] if [=mobile=] is false. User Agents MUST return the empty string for [=model=] even if [=mobile=] is true, except on platforms where the model is typically exposed." Yes, it's ok to have a MUST with a large exception.

> [=User agents=] MAY return the empty string or a fictitious value for [=platform architecture=] 
> or [=model=], and MAY return the [=significant version=] value for [=full version=] in case they consider that exposing the real values increases their users' privacy risk beyond the user benefit of the use-cases they enable."

More clear to write these as separate sentences rather than linked by "and", the sentence ends up too complex IMO. The reason to expose a fictitious value would more likely be compatibility than privacy. I think it's clearer to make the MAY unconditional instead of stating a specific fuzzy reason. It's hard to see how UA authors could even make the tradeoff you described, since the spec doesn't describe either the privacy risk or the user benefit.

More generally, I don't think caring about privacy should be optional but it's better than nothing. 

> > Additional comment: I think we should drop "full version" entirely unless there is at least one browser that wants to expose it. Safari would not do so. I guess it sounds like Chrome would indeed want it?

> Chrome would want to expose it in order to match the current capabilities of the User-Agent string. Maybe over time we can find a better alternative to tackle the use cases it enables. If so, we could deprecate and remove it then.

The only use case there that seems to call for it is vulnerability filtering. Is that a real thing? Which websites filter browsers based on detailed version for security reasons? When Safari dropped the detailed version from its UA string, we heard many concerns raised, but not this one, so I am curious.

mcatanzaro:
> I don't think "where binary downloads are unlikely" is sufficient. It ought to be be left blank on platforms where different binary downloads for different CPUs are unlikely.

I would use MAY (or perhaps SHOULD) instead of MUST here. Let the browser choose whether or not to expose CPU architecture. I agree it only makes sense on Windows.

index.bs
```diff
@@ -170,22 +170,41 @@ User Agent Hints {#http-ua-hints}
...
+      architecture (for example: "ARM64", or "ia32")
+*   <dfn for="user agent" export>model</dfn> - The [=user agent=]'s device model (for example: "",
+      or "Pixel 2 XL")
+*   <dfn for="user agent" export>mobileness</dfn> - A boolean indicating if the [=user agent=]'s
+      device is a mobile device. (for example: ?0 or ?1)
 
 [=User agents=] SHOULD keep these strings short and to the point, but servers MUST accept arbitrary
 values for each, as they are all values constructed at the [=user agent=]'s whim.
 
+[=User agents=] SHOULD return the empty string or a fictitious value for [=platform architecture=]
```
yoavweiss, line 197: @othermaciej - I rewrote the conditions here. PTAL?

yoavweiss: FWIW, I believe that for CPU arch, those 2 conditions are also met on Android and Linux (and not just Windows), despite the OS' attempt to centralize app installs. On macOS, binary downloads are similarly likely, but the support for fat binaries makes it so that the second condition is not met (as you implied in an earlier comment).

yoavweiss: @bslassey tells me that arch is likely to matter very little on Android for binary downloads, so the second condition is not likely to hold there as well.

yoavweiss:
@othermaciej - Friendly ping! :) Can you take another look?

index.bs
othermaciej: This generally seems ok.

index.bs
```diff
@@ -170,22 +170,41 @@ User Agent Hints {#http-ua-hints}
...
+unless the user's platform is one where both the following conditions apply:
+ * Binary download of executables is likely.
+ * Different CPU architectures are likely to require different binary executable resources.
```
othermaciej, line 200: This is a bit different than the second condition I had in mind, which is that binary downloads for more than one different CPU architectures are likely to be available. So for example, if binary downloads are only commonly found for one CPU architecture, even though others technically exist, that wouldn't meet the condition I was thinking of. I raise this only for consideration. This would be sufficient.

index.bs
```diff
@@ -170,22 +170,41 @@ User Agent Hints {#http-ua-hints}
...
+
+[=User agents=] MUST return the empty string for [=model=] if [=mobileness=] is false or on
+platforms where the model is not typically exposed.
+
+[=User agents=] MAY return the empty string or a fictitious value for [=full version=], [=platform
+architecture=] or [=model=] in cases they consider that exposing that value increase their user's
+privacy risk beyond the user benefit of the use-cases they enable.
```
othermaciej, line 207: Copying feedback from top-level comment:

The reason to expose a fictitious value would more likely be compatibility than privacy. I think it's clearer to make the MAY unconditional instead of stating a specific fuzzy reason. It's hard to see how UA authors could even make the tradeoff you described, since the spec doesn't describe either the privacy risk or the user benefit.

More generally, I don't think caring about privacy should be optional but it's better than nothing.

I'd strongly suggest dropping everything from "in cases" onward, but if you want some guidance for the MAY, I'd replace it with "for privacy, compatibility, or other reasons."

yoavweiss: Changed to your wording.

FWIW, I wasn't trying to convey that caring for privacy is optional, but that there's a tradeoff here between usability and privacy which should be taken into consideration.

index.bs
```diff
@@ -170,22 +170,41 @@ User Agent Hints {#http-ua-hints}
...
 [=User agents=] SHOULD keep these strings short and to the point, but servers MUST accept arbitrary
 values for each, as they are all values constructed at the [=user agent=]'s whim.
 
+[=User agents=] SHOULD return the empty string or a fictitious value for [=platform architecture=]
+unless the user's platform is one where both the following conditions apply:
+ * Binary download of executables is likely.
+ * Different CPU architectures are likely to require different binary executable resources.
+
+[=User agents=] MUST return the empty string for [=model=] if [=mobileness=] is false or on
+platforms where the model is not typically exposed.
```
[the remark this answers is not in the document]
yoavweiss: Happy to change it if you think it's clearer.

index.bs
yoavweiss: That's fair. Adding "likely to be available" to the condition.

yoavweiss:
@othermaciej - Changed based on your feedback. PTAL?

yoavweiss:
Thanks for reviewing and for the great feedback! :)

Issue #89, opened 2020-03-16
Add bug workaround use-case for significant browser version


yoavweiss:

yoavweiss:
Closed by #92

PR #92, opened 2020-03-18
Add download and bug workaround use cases

### LINKED ISSUE #88: Add binary download use-case for architecture
As mentioned in #58, the use-case for architecture is not covered in the [use-cases](https://github.com/WICG/ua-client-hints#use-cases). It should be.

yoavweiss:
Closes #88 and #89 

@morlovich - can you review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,9 @@ that the extent of a site's usage can be monitored and evaluated.
 This section attempts to document the current uses for the `User-Agent` string,
 and how similar functionality could be enabled using UA-CH.
 
-## Differential serving based on browser features
+## Differential serving
+
+### Based on browser features
 This use-case enables services like [polyfill.io](https://polyfill.io) to serve
 custom-tailored polyfills to their users, without bloating up the experience of
 modern browser users.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -238,9 +240,23 @@ meaningful version, and map that to a list of available features. That enables
...
+other issues, and those bugs are by definition not something that can be
+feature detected. Therefore, content needs to avoid them altogether for
+affected browser versions.
+
+For that use case, servers need to be aware of the browser and its meaningful
+version, be aware of browser bugs that impact them, and apply workarounds if
+the current browser version is impacted.
+
+Services that wish to do that using UA-CH will need to inspect the `Sec-CH-UA`
+header, sent by default on every request, and use it to modify their response.
```
morlovich: My question upon reading this is that those sorts of bugs are often fixed within a lifetime of a single "meaningful" version.

README.md
```diff
@@ -364,6 +380,17 @@ Since such messaging doesn't require any server-side adaptation, it's better
...
+## Download of appropriate binary executables
+Some sites are used to download binary executables of native applications, and
+need to be able to propose the right binary to the user by default.  The right
+binary executable for the current user depends on a few factors: their
+operating system, its version, as well as their CPU architecture.
+
+In order to tackle that use case, download sites can opt-in to receive the
+`UA-Platform`, `UA-Platform-Version` and `UA-Architecture` hints (or query them
+through the API), in order to ensure the right binary is offered to the user by
+default.
```
morlovich: Non-seriously: request UA-Platform-LinuxPackageFormat

README.md
yoavweiss: That's not necessarily true. E.g. the recent `same-site` issues that require User-Agent sniffing to work around old versions of Safari, WebP issues in older versions of Android, etc.

README.md
yoavweiss: There's certainly a use-case there, but since it's not something the current UA supports, I think it's something we can think about further down the line, if there's demand

yoavweiss:
@morlovich - friendly ping! :)

morlovich:
LGTM? Is that how it works here?

yoavweiss:
> LGTM? Is that how it works here?

Yup! :) (as well as approving the PR, but that's not strictly needed here)

PR #95, opened 2020-04-08
Remove `brand` reference from GREASE example

### LINKED ISSUE #21: Set of UA strings needs more explanation

yoavweiss:
Closes #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,7 +481,7 @@ Let's examine a few examples:
 * Browsers based on Chromium may use a similar UA string, but use their own brand as part of the set, enabling sites to count them.
   - `"Chrome"; v="73", "Awesome Browser"; v="60", "Chromium"; v="73"`
 
-We'd reflect this value in the `NavigatorUAData.userAgentData.uaList` attribute, which returns an array of dictionaries containing brand and version.
+We'd reflect this value in the `navigator.userAgentData.uaList` attribute, which returns an array of dictionaries containing brand and version.
 
 [4]: https://tools.ietf.org/html/draft-ietf-tls-grease-01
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -481,7 +481,7 @@ Let's examine a few examples:
 * Browsers based on Chromium may use a similar UA string, but use their own brand as part of the set, enabling sites to count them.
   - `"Chrome"; v="73", "Awesome Browser"; v="60", "Chromium"; v="73"`
 
-We'd reflect this value in the `NavigatorUAData.brand` attribute, which returns an array of dictionaries containing brand and version.
+We'd reflect this value in the `NavigatorUAData.userAgentData.uaList` attribute, which returns an array of dictionaries containing brand and version.
```
domenic, line 484:
```suggestion
We'd reflect this value in the `navigator.userAgentData.uaList` attribute, which returns an array of dictionaries containing brand and version.
```

PR #101, opened 2020-04-27
Queue a task on the permissions task source

### LINKED ISSUE #77: What task queue should the Promise be queued on?
In [getEntropyValues](https://wicg.github.io/ua-client-hints/#getHighEntropyValues), we're queueing a Promise to be resolved, but it's unclear on what task queue that should happen. We need to figure out.

/cc @domenic @bzbarsky

yoavweiss:
Closes #77

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/ua-client-hints/pull/101.html" title="Last updated on Apr 28, 2020, 8:59 PM UTC (6532025)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/101/7911817...yoavweiss:6532025.html" title="Last updated on Apr 28, 2020, 8:59 PM UTC (6532025)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -398,9 +398,9 @@ On getting, the {{NavigatorUAData/mobile}} attribute must return the [=user agen
 
 The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></dfn> method MUST run these steps:
 
-1. Let |p| be a [=a new promise=] created with {{NavigatorUAData}}'s [=relevant realm=].
+1. Let |p| be a [=a new promise=] created with [=current settings object=]'s [=environment settings object/realm=].
 
-2. If [=this=]'s [=relevant global object=] is not identical to {{NavigatorUAData}}'s [=relevant global object=], then:
+2. If {{NavigatorUAData}}'s [=relevant global object=] is a [=Window=] and its [=associated Document=] is [=fully active=], then:
 
     1. [=Reject=] |p| with a {{WrongDocumentError}}.
 
--- a/index.bs
+++ b/index.bs
@@ -23,7 +23,6 @@ spec:fetch; type:dfn; for:/; text:request
 spec:webidl; type:dfn; text:resolve
 spec:infra; type:dfn; text:user agent
 spec:infra; type:dfn; for:/; text:list
-spec:permissions; type:dfn; for:/; text: permission task source
 </pre>
 <pre class="anchors">
 urlPrefix: https://tools.ietf.org/html/draft-ietf-httpbis-header-structure; spec: I-D.ietf-httpbis-header-structure
@@ -399,9 +398,15 @@ On getting, the {{NavigatorUAData/mobile}} attribute must return the [=user agen
 
 The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></dfn> method MUST run these steps:
 
-1. Let |p| be a [=a new promise=].
+1. Let |p| be a [=a new promise=] created with {{NavigatorUAData}}'s [=relevant realm=].
 
-2.  Run the following steps [=in parallel=]:
+2. If [=this=]'s [=relevant global object=] is not identical to {{NavigatorUAData}}'s [=relevant global object=], then:
+
+    1. [=Reject=] |p| with a {{ReferenceError}}.
+
+    2. Return |p|.
+
+3.  Run the following steps [=in parallel=]:
 
     1. Let |uaData| be a new {{UADataValues}}.
 
@@ -417,9 +422,7 @@ The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></df
 
     7. [=Queue a task=] on the [=permission task source=] to [=resolve=] |p| with |uaData|.
 
-ISSUE: Add a specific task source this is queued on.
-
-3.  Return |p|.
+4.  Return |p|.
 
 Security and Privacy Considerations {#security-privacy}
 ===================================
--- a/index.bs
+++ b/index.bs
@@ -22,6 +22,8 @@ Boilerplate: omit conformance, omit feedback-header
 spec:fetch; type:dfn; for:/; text:request
 spec:webidl; type:dfn; text:resolve
 spec:infra; type:dfn; text:user agent
+spec:infra; type:dfn; for:/; text:list
+spec:permissions; type:dfn; for:/; text: permission task source
 </pre>
 <pre class="anchors">
 urlPrefix: https://tools.ietf.org/html/draft-ietf-httpbis-header-structure; spec: I-D.ietf-httpbis-header-structure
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -22,6 +22,8 @@ Boilerplate: omit conformance, omit feedback-header
 spec:fetch; type:dfn; for:/; text:request
 spec:webidl; type:dfn; text:resolve
 spec:infra; type:dfn; text:user agent
+spec:infra; type:dfn; for:/; text:list
+spec:permissions; type:dfn; for:/; text: permission task source
```
domenic, line 26: This should not be necessary. I guess the issue is that the permissions spec is not exporting this definition? If so, then maybe we shouldn't use it, instead of using Bikeshed features to bypass that protection.

I'd suggest filing an issue/PR on the permissions spec to discuss exporting this task source.

yoavweiss: Filed https://github.com/w3c/permissions/issues/206

yoavweiss: Done

yoavweiss:
Also added a few changes to the processing model here based on feedback from @dtapuska on the relevant [Chromium CL](https://chromium-review.googlesource.com/c/chromium/src/+/2166103)

index.bs
```diff
@@ -398,9 +398,15 @@ On getting, the {{NavigatorUAData/mobile}} attribute must return the [=user agen
 
 The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></dfn> method MUST run these steps:
 
-1. Let |p| be a [=a new promise=].
+1. Let |p| be a [=a new promise=] created with {{NavigatorUAData}}'s [=relevant realm=].
 
-2.  Run the following steps [=in parallel=]:
+2. If [=this=]'s [=relevant global object=] is not identical to {{NavigatorUAData}}'s [=relevant global object=], then:
```
yoavweiss, line 403: @dtapuska - Is this a good equivalent to `!GetExecutionContext()`?

dtapuska: + @domenic

I don't think this is quite right. Isn't this the same as NavigatorUAData?  I think what you mean is that the RelevantRealm != IncumbentRealm then throw WrongDocument ?

domenic: Incumbent realm should not be used by new specs.

If you are trying to make sure that the document is alive, then you need to check if the relevant global object is a Window and its associated Document is fully active.

What are you actually trying to express here, in terms of developer-facing benefit, not in terms of Chromium implementation details?

yoavweiss: > If you are trying to make sure that the document is alive, then you need to check if the relevant global object is a Window and its associated Document is fully active.


Makes sense, thanks!

> What are you actually trying to express here, in terms of developer-facing benefit, not in terms of Chromium implementation details?

I'm trying to express that detached frames should reject here.
The developer-facing benefit is reduced confusion as the rest of UA data returns nullyfied values in that case.

domenic: > the rest of UA data returns nullyfied values in that case.

That doesn't appear to be true, at least for anything in https://html.spec.whatwg.org/multipage/system-state.html#client-identification .

domenic: In general I'd recommend against taking a dependency on global state in this way, and instead have these values be independent of such external factors.

yoavweiss: > That doesn't appear to be true, at least for anything in https://html.spec.whatwg.org/multipage/system-state.html#client-identification .

The plot thickens...
Chromium has a [test](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/fast/dom/Window/property-access-on-cached-properties-after-frame-removed.html?q=fast%2Fdom%2FWindow%2Fproperty-access-on-cached-properties-after-frame-removed.html&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2F) that enforces this.

I wrote a small [test page](https://yoav-scratchpad.glitch.me/ua_detached.html) for `navigator.userAgent` and it seems like Safari and Chromium both nullify the string, while Firefox does not...

domenic: Ugh, that's bizarre. I hope we can align Chromium with the spec? There's just no reason to make these getters dependent on state like that.

yoavweiss: Maybe the easiest path forward would be to downsize this PR to the task queueing only, drop the Promise rejection, find a different ExecutionContext to queue the task on in the Chromium implementation (from the `ScriptState`) and open an issue on the HTML spec to figure out the nullification of attributes on detached documents.

domenic: That makes sense to me.

Note that the spec as written rejects the promise but the rejection will never be seen, since the event loop is inactive for those documents.

yoavweiss: Removed the promise rejection, but left the realm on which the promise is created.

> Note that the spec as written rejects the promise but the rejection will never be seen, since the event loop is inactive for those documents.

Is this because of the realm the promise is created with?

domenic: No, it's because the event loop just doesn't run for inactive documents.

index.bs
```diff
@@ -398,7 +398,7 @@ On getting, the {{NavigatorUAData/mobile}} attribute must return the [=user agen
 
 The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></dfn> method MUST run these steps:
 
-1. Let |p| be a [=a new promise=].
+1. Let |p| be a [=a new promise=] created with [=current settings object=]'s [=environment settings object/realm=].
```
domenic, line 401:
```suggestion
1. Let |p| be a [=a new promise=] created in [=the current Realm=].
```

This may necessitate adding an anchors block entry for https://tc39.es/ecma262/#current-realm since ES spec/Bikeshed integration is still sad.

index.bs
```diff
@@ -398,7 +398,7 @@ On getting, the {{NavigatorUAData/mobile}} attribute must return the [=user agen
 
 The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></dfn> method MUST run these steps:
 
-1. Let |p| be a [=a new promise=].
+1. Let |p| be a [=a new promise=] created with <a href="https://tc39.es/ecma262/#current-realm">the current realm</a>.
```
domenic, line 401:
```suggestion
1. Let |p| be a [=a new promise=] created in <a href="https://tc39.es/ecma262/#current-realm">the current realm</a>.
```

index.bs
yoavweiss: Added a direct link, so that this won't be blocking. Happy to also file a PR for adding that to ES, but would appreciate some pointers

domenic: Nah, the ES spec won't accept PRs to add Bikeshed-isms, sadly. You have to work around it :(.

A direct link is reasonable, although if you're going to link somewhere more than once you could use the anchors block instead: cf. https://github.com/whatwg/streams/blob/webidl-rewrite/index.bs#L37

yoavweiss: Makes sense, thanks! :)

Issue #105, opened 2020-05-04
Client Hint for CPU "bitness"


yoavweiss:
There may be use-cases in which detailed CPU architectures is needed (e.g. download of highly-optimized binary executables).
It would be good to keep the current `architecture`/`Sec-CH-UA-Arch` relatively low-entropy (for the common case), and potentially expose more detailed info only to sites that opt-in to that specifically.

othermaciej:
I don't think this is a good idea; it's quite a bit of extra entropy. Is this exposed in any current User-agent strings?

yoavweiss:
> I don't think this is a good idea

I'm not 100% sure it is either, therefore I marked it as "enhancement", that may be part of a future iteration. 

The motivation here: As part of https://github.com/WICG/ua-client-hints/pull/106 we significantly reduce the entropy currently exposed (at least in Chromium, at least in some platforms), under the assumption that it's sufficient for binary downloads.

The idea is that *if* legitimate use-cases arise that actually need that extra information, we'd be able to provide that information without forcing the more common use cases to touch that entropy.

bslassey:
@othermaciej this is related to https://github.com/WICG/ua-client-hints/pull/106#discussion_r421971178 where you said that the 32/64bit distinction is important. Entropy-wise, having a single hint that has arm32, arm64, x86 and x86_64 versus two hints with arm, x86 and 32, 64 values is the same. 

Practically, I see some benefits from separating the two. First, it is forward compatible in that as the architectures have expanded (i.e. when 32 bit and 64 bit chips were introduced) existing binaries were backward compatible (so 16bit binaries ran on 32 bit chips and 32 bit binaries ran on 64 bit chips). In the future when we have 128 bit chips, I would hate to see a world where the x86_128 or arm128 tokens caused sites serving 64 bit binaries to error out. 

On a similar token, right now if I am a site hosting 32 bit x86 or 32 bit arm binaries, I don't particularly care about the distinction of 32 and 64 bit when user is trying to download them.

othermaciej:
Not all 64-bit operating systems for arm or x86 are able to run 32-bit binaries. (For example, the latest versions of macOS and iOS are strictly 64-bit only; and there's apparently Linux distributions that have 64-bit only userland). But perhaps this is not the common case.

[...]

PR #106, opened 2020-05-04
Define values for CPU architecture

### LINKED ISSUE #61: Should Sec-UA-Arch define canonical names for popular architectures, or have a registry?
In current user agent strings, it's hard to make use of CPU architecture without reference to platform.

For example, popular browser UA strings refer to the x86-64 architecture as `x86_64` on Linux, `x64` on Windows, and `Intel` on Macs. To make it reasonable and useful to request `Sec-UA-Arch` by itself, there probably needs to be a consistent name for any given CPU architecture. To achieve this, the spec either needs to predefine some, or point to a registry.

You could just say that browsers ought to use whatever architecture name they send in the UA string, and clients just have to deal with multiple synonyms for the same architecture, but that misses a potential one-time opportunity for a useful cleanup.

(Note though that in #58 I suggest dropping this field, or allowing it to be blank or fictional, for browsers that chose not to expose the true CPU architecture at all.)

yoavweiss:
Closes #61 

/cc @othermaciej

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/ua-client-hints/pull/106.html" title="Last updated on May 19, 2020, 7:00 AM UTC (7b6d4c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/106/facbb0e...yoavweiss:7b6d4c6.html" title="Last updated on May 19, 2020, 7:00 AM UTC (7b6d4c6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -191,7 +191,7 @@ values for each, as they are all values constructed at the [=user agent=]'s whim
 * 32 bit x86 CPU architectures => "x86"
 * 64 bit x86 CPU architectures => "x86_64"
 * 32 bit ARM CPU architectures => "arm"
-* 64 bit ARM CPU architectures => "arm_64"
+* 64 bit ARM CPU architectures => "arm64"
 
 Other CPU architectures could be mapped into one of these values in case that makes sense, or be
 mapped to the empty string.
--- a/index.bs
+++ b/index.bs
@@ -188,17 +188,16 @@ values for each, as they are all values constructed at the [=user agent=]'s whim
 
 [=User agents=] MUST map higher-entropy [=platform architecture=] values to the following buckets:
 
-* 32 bit x86 CPU architectures => "x86"
-* 64 bit x86 CPU architectures => "x86_64"
-* 32 bit ARM CPU architectures => "arm"
-* 64 bit ARM CPU architectures => "arm64"
+* x86 CPU architectures => "x86"
+* ARM CPU architectures => "arm"
 
 Other CPU architectures could be mapped into one of these values in case that makes sense, or be
 mapped to the empty string.
 
-ISSUE 105: There might be use-cases for higher-entropy and more specific CPU architectures (e.g.
-for the download of highly optimized executable binaries). If necessarrry, we could support those
-use-cases through one or more separate hints.
+ISSUE 105: There might be use-cases for higher-entropy, more specific CPU architectures (e.g. 32
+vs. 64 bit architectures, or specific instruction sets for the download of highly optimized
+executable binaries). If necessary, we could support those use-cases through one or more separate
+hints.
 
 
 The 'Sec-CH-UA-Arch' Header Field {#sec-ch-arch}
--- a/index.bs
+++ b/index.bs
@@ -179,12 +179,24 @@ a number of properties for itself:
 *   <dfn for="user agent" export>full version</dfn> (for example: "72.0.3245.12", "3.14159", or "297.70E04154A")
 *   <dfn for="user agent" export>platform brand</dfn> (for example: "Windows", "iOS", or "AmazingOS")
 *   <dfn for="user agent" export>platform version</dfn> (for example: "NT 6.0", "15", or "17G")
-*   <dfn for="user agent" export>platform architecture</dfn> (for example: "ARM64", or "ia32")
+*   <dfn for="user agent" export>platform architecture</dfn> (for example: "arm", or "x86")
 *   <dfn for="user agent" export>model</dfn> (for example: "", or "Pixel 2 XL")
 *   <dfn for="user agent" export>mobileness</dfn> (for example: ?0 or ?1)
 
 [=User agents=] SHOULD keep these strings short and to the point, but servers MUST accept arbitrary
 values for each, as they are all values constructed at the [=user agent=]'s whim.
+
+[=User agents=] MUST map higher-entropy [=platform architecture=] values to the following buckets:
+
+* x86 CPU architectures => "x86"
+* ARM CPU architectures => "arm"
+
+Other CPU architectures could be mapped into one of these values in case that makes sense, or be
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -179,13 +179,25 @@ a number of properties for itself:
...
 *   <dfn for="user agent" export>model</dfn> (for example: "", or "Pixel 2 XL")
 *   <dfn for="user agent" export>mobileness</dfn> (for example: ?0 or ?1)
 
 [=User agents=] SHOULD keep these strings short and to the point, but servers MUST accept arbitrary
 values for each, as they are all values constructed at the [=user agent=]'s whim.
 
+[=User agents=] MUST map higher-entropy [=platform architecture=] values to the following buckets:
+
+* x86 CPU architectures => "x86"
+* ARM CPU architectures => "arm"
```
othermaciej, line 192: I did ask for less of this info, but I think this is so much less that it's not useful any more. In practice, x86_64 and arm64 are different enough architectures that in practice they cannot be considered the same as their 32-bit cousins.

What value does this info provide that the empty string does not?

Not only does this not directly provide value, but I also I think there's a hazard here that x86 vs arm could start to be wrongly used as a desktop vs mobile distinction. Yes, there is a token for this purpose, but I'm hard pressed to think of any other more valid use of a CPU architecture field with only these two values.

index.bs
```diff
@@ -179,13 +179,25 @@ a number of properties for itself:
...
+
+Other CPU architectures could be mapped into one of these values in case that makes sense, or be
+mapped to the empty string.
```
othermaciej, line 195: For Apple platforms, we'd like the freedom to map even x86-family or arm-family architectures to the empty string, or to map an architecture to the "wrong" family when convenient.  It's not clear if this spec language is meant to allow that. I think this should be explicitly permitted, if this two-value field remains.

yoavweiss: This would be explicitly permitted by https://github.com/WICG/ua-client-hints/pull/90/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR205

index.bs
yoavweiss: The general thinking when talking to @bslassey about this was that this would enable 32 bit installers that then can then download the appropriate variant, as a bare minimum, and we could later maybe enable a more accurate CPU reporting if needed. (which would include both the 32/64 bit distinction, as well as potentially other characteristics, like ARM version)

I agree that the 32/64 bit distinction feels more useful than the latter part, so I'm happy to modify the PR to include it. In browsers that would take the Privacy Budget route, it would increase the "price" of this API, but that's probably fine.

index.bs
```diff
@@ -179,13 +179,27 @@ a number of properties for itself:
...
 
 [=User agents=] SHOULD keep these strings short and to the point, but servers MUST accept arbitrary
 values for each, as they are all values constructed at the [=user agent=]'s whim.
 
+[=User agents=] MUST map higher-entropy [=platform architecture=] values to the following buckets:
+
+* 32 bit x86 CPU architectures => "x86"
+* 64 bit x86 CPU architectures => "x86_64"
+* 32 bit ARM CPU architectures => "arm"
+* 64 bit ARM CPU architectures => "arm_64"
```
othermaciej, line 194: This architecture is conventionally called arm64, not arm_64. (Unlike the 64-bit version of x86, which is known by many names, where x86_64 is one of the common ones.

index.bs
othermaciej: Sounds like it would. Might need some re-review once both PRs land.

index.bs
yoavweiss: Yeah, I guess we could stick with current conventions. Changed it.

index.bs
```diff
@@ -188,15 +188,17 @@ values for each, as they are all values constructed at the [=user agent=]'s whim
...
 
 Other CPU architectures could be mapped into one of these values in case that makes sense, or be
 mapped to the empty string.
 
 ISSUE 105: There might be use-cases for higher-entropy and more specific CPU architectures (e.g.
-"bitness" or for the download of highly optimized executable binaries). If necessarrry, we could
-support those use-cases through one or more separate hints.
+for the download of highly optimized executable binaries). If necessarrry, we could support those
```
morlovich, line 200: Typo, necessarrry.
(I think your 'r' key may be misbehaving, it's a second of these I noticed..)

yoavweiss: indeed... (as well as me 'e', and occasionally my 't'). COVID-19 is holding up refreshes...

yoavweiss:
@othermaciej Based on https://github.com/WICG/ua-client-hints/issues/105#issuecomment-626904363, it seems like there's agreement on 2 values (without "bitness") and that bitness would be something we may tackle in the future as necessary. I reverted the values back to "arm" and "x86".

PTAL?

index.bs
tresf: >  I also I think there's a hazard here that x86 vs arm could start to be wrongly used as a desktop vs mobile distinction.

I'm not sure who would make this assumption.  The Surface Pro X ships with a native ARM version of Edge.  Rumors say other software giants (😉) may soon offer the same.  Chromebooks can run Linux binaries, but they're architecture dependant.  ISVs may want metrics as well.

I'm drafting this into a project for my own analytics (future proofing, anticipating the analytics platforms to slowly deprecate the old in favor of the new) and find the directives hard to follow.  For example, current I convert `amd64` to `x86_64` for the User-Agent string to mimic other systems, but this seems like I should simply use `x86`, which seems like a loss of information.

yoavweiss: Apologies but I missed your comment. x86 is indeed a loss of information, but that loss is intentional, to enable folks to access only the partial information when it is enough. The hint discussed in https://github.com/WICG/ua-client-hints/issues/105 will enable accessing bitness as well.

PR #107, opened 2020-05-08
Make GREASE a MUST

### LINKED ISSUE #60: GREASE-like UA strings proposal might not work unless the MAY becomes a MUST
> History has shown us that there are real incentives for user agents to lie about their branding in order to thread the needle of sites' sniffing scripts, and prevent their users from being blocked by UA-based allow/block lists. .... Reseting expectations may help to prevent abuse of the UA string’s brand in the short term, but probably won’t help in the long run. ... User agents SHOULD model UA as a set, rather than a single entry. They then could encourage standardized processing of the UA string, by randomly including additional, intentionally incorrect, comma-separated entries with arbitrary ordering. 

This seems to be a good idea to maintain the viability of spoofing without totally hiding real branding. However, that's probably not the case if it's optional. A very popular browser could make itself harder to spoof by not following this.

yoavweiss:
Closes #60 

/cc @othermaciej

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/ua-client-hints/pull/107.html" title="Last updated on May 8, 2020, 10:35 PM UTC (77b46a1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/107/482a05b...yoavweiss:77b46a1.html" title="Last updated on May 8, 2020, 10:35 PM UTC (77b46a1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -498,17 +498,24 @@ Let's examine a few examples:
     set, enabling sites to count them.
     - `"Chrome"; v="73", "Awesome Browser"; v="60", "Chromium"; v="73"`
 
+
+[=User agents=] MUST include more than a single value in [=user agent/brands=], where at least one
+of these values is an arbitrary value.
+
+When adding arbitrary values to [=user agent/brands=], [=user agents=] MUST make sure that receivers
+of the header adhere to [=Structured Header=] parsing, by adding escaped double-quotes, commas and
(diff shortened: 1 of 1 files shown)


yoavweiss:
Thanks for reviewing! :)

PR #124, opened 2020-08-31
Incorporates experiment feedback and review from interested users and other stakeholders.

jwrosewell:
Following review and testing of the provided experiments with interested users and other stakeholders this PR modifies the document to achieve the following.

1. Avoid conflating the specification for User Agent Client-Hints with;
a. the migration of established methods of transferring the same information; and
b. identifying and resolving all privacy issues via this specification and a limited number of related specifications.

2. References a broader set of informative work including the establishment of preference choices at the point of user agent upgrade and installation, default options and the idea of a Privacy Budget.

3. Removes GREASE as it makes the information exposed in the proposed headers harder to consume in practice which goes against the primary purpose of the specification.

4. Expanded the introduction and range of use cases for the information Client-Hints seek to control.

5. Added Issues to the document raised in other forums.

6. Considers the need to restructure the existing User-Agent field value to achieve similar goals and compare the two proposals.

***
<a href="https://pr-preview.s3.amazonaws.com/jwrosewell/ua-client-hints/pull/124.html" title="Last updated on Sep 1, 2020, 8:47 AM UTC (2598594)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/124/dbb11ce...jwrosewell:2598594.html" title="Last updated on Sep 1, 2020, 8:47 AM UTC (2598594)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -97,6 +97,16 @@ urlPrefix: https://tools.ietf.org/html/draft-ietf-httpbis-header-structure; spec
     "href": "https://wicg.github.io/client-hints-infrastructure/",
     "title": "Client Hints Infrastructure",
     "author": [ "Yoav Weiss" ]
+  },
+  "CMA-UK": {
+    "href": "https://www.gov.uk/cma-cases/online-platforms-and-digital-advertising-market-study",
+    "title": "Online platforms and digital advertising market study",
+    "author": [ "UK Competition and Markets Authority" ]
+  },
+  "Privacy-Budget": {
+    "href": "https://github.com/bslassey/privacy-budget",
+    "title": "Combating Fingerprinting with a Privacy Budget",
+    "author": [ "Brad Lassey" ]
   }
 }
 </pre>
@@ -105,7 +115,7 @@ Dependencies {#dependencies}
 ============
 
 This document is dependent on the following related specifications which the reader must be 
-familar with before reading this document.
+familiar with before reading this document.
 
 1.  Experimental IETF specification for Client Hints ([[I-D.draft-ietf-httpbis-client-hints]]) 
     which expires in January 2021.
@@ -131,12 +141,12 @@ When combined with other information the `User-Agent` field value could be used 
 user agents and therefore might represent a risk to people's privacy. This problem is most severe 
 when content is served over insecure HTTP where access providers can view the information. Privacy
 concerns are mitigated to some extent when secure protocols such as HTTPS are used. Only parties 
-known to the user or the organisation whose domain is displayed in the address bar of the user 
+known to the user or the organization whose domain is displayed in the address bar of the user 
 agent will have access to the `User-Agent` field value.
 
 [[RFC7231]] does not mandate the structure of the information contained in the field value. It has
 evolved over several decades by multiple browser vendors and often contains information that is not
-needed. Robust solutions ([[51Degrees]] [[DeviceAtlas]], [[Scientiamobile]]) for parsing 
+needed. Robust solutions ([[51Degrees]], [[DeviceAtlas]], [[Scientiamobile]]) for parsing 
 `User-Agent` field values know which segments to ignore or focus on and perform with over 99% 
 accuracy in practice.
 
@@ -172,8 +182,8 @@ Use cases for the information contained in `User-Agent` field values include:
     application on Android in order to match each platform's aesthetic and design patterns.
 
 3.  Similarly to #1, OS revisions and architecture can be responsible for specific bugs which can
-    be worked around in website's code, and useful for things like selecting appropriate executables
-    for download (32 vs 64 bit, ARM vs Intel, etc).
+    be worked around in website's code, and useful for things like selecting the appropriate 
+    executable for download (32 vs 64 bit, ARM vs Intel, etc).
 
 4.  Sophisticated developers use model/make to tailor their sites to the capabilities of the
     device (e.g. [[FacebookYearClass]]) and to pinpoint performance bugs and regressions which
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -74,20 +67,88 @@ urlPrefix: https://tools.ietf.org/html/draft-ietf-httpbis-header-structure; spec
...
 
+Dependencies {#dependencies}
+============
+
+This document is dependent on the following related specifications which the reader must be 
+familiar with before reading this document.
+
+1.  Experimental IETF specification for Client Hints ([[I-D.draft-ietf-httpbis-client-hints]]) 
+    which expires in January 2021.
+
```
reschke, line 122: That's misleading. The draft has been approved by the IESG and is in the RFC Publication Queue.

reschke (changes requested):

index.bs
jwrosewell: @reschke could you provide some links and more information? The edit was based on the status of the document referenced originally.

reschke: Links: <https://datatracker.ietf.org/doc/draft-ietf-httpbis-client-hints/> and <https://www.rfc-editor.org/current_queue.php#draft-ietf-httpbis-client-hints>

The *draft* itself does indeed expire 6 months after publication, but that really is irrelevant; in particular because it will be published as RFC before that time.

jwrosewell: I've reviewed the [RFC2026](https://tools.ietf.org/html/rfc2026#section-4.2.1), and its amendments, and do not believe this dependent specification is on any standards track within the IETF. The definition of Experimental status is as follows.

>  The "Experimental" designation typically denotes a specification that
   is part of some research or development effort.  Such a specification
   is published for the general information of the Internet technical
   community and as an archival record of the work, subject only to
   editorial considerations and to verification that there has been
   adequate coordination with the standards process (see below).  An
   Experimental specification may be the output of an organized Internet
   research effort (e.g., a Research Group of the IRTF), an IETF Working
   Group, or it may be an individual contribution.

The pull request has been amended to reference RFC2026 and highlight the experimental nature of the Client Hints document.

yoavweiss:
Thank you for your interest.
I think it might be better to split up the different points you want to discuss into separate issues, so that we can review each of them individually.
For example, including use-cases explored in the explainer into the specification seems non-controversial.
The rest seems to warrant more discussion.

index.bs
reschke: ack

jwrosewell:
@yoavweiss would you prefer we did this by raising issues against the repo for the major points and then reference this pull request for the proposed text changes?

If so then it would be helpful to get more guidance on the other points? For example; the feedback separates the technical specification of a new feature from deployment and migration concerns which would not be included in a W3C or HTTP WG recommendation or standards track document. This seems non-controversial.

We assume Google would not progress the mass deployment of such a major feature and related change without advancing the technical specifications involved to agreement within the W3C and IETF.

yoavweiss:
I prefer to close the above PR as it mixes many different subjects and as such, is extremely hard to review. The best route would be to open separate issues, discuss them, and send separate PRs as necessary.

jwrosewell:
The summary issues requested have been opened and linked back to this pull request.

Perhaps the October TPAC could be used to discuss these issues via a conference call?

In anycase I trust you and Google will agree it is important to address these matters before mass deployment or deprecation of an existing de facto standard of interoperability.

jwrosewell:
As today is the last day to submit TPAC Breakout session proposals I've added a [place holder session](https://www.w3.org/wiki/TPAC/2020/SessionIdeas#User_Agent_Client_Hints).

yoavweiss:
TBH, that doesn't seem like the best use of time for a breakout session. There seem to be 45 different sessions competing over 5 slots. Might be best to dedicate that time to high-level subjects of broader interest, and reserve issue resolution to GitHub.

/cc @dontcallmedom for his opinions

jwrosewell:
Perhaps WICG has session(s) similar to PING and the Improving Web Advertising BG which could be used in place of the Breakout sessions?

yoavweiss:
Not as a CG. There were a few incubations that ran meetings last week, but UA-CH was not one of them.

dontcallmedom:
I would tend to agree with @yoavweiss that the issue resolution isn't what most people would expect at the unconference (which may also imply few people would show up given the competition for attention).

Let me know ASAP if the session description needs to be updated

jwrosewell:
Thank you @travisleithead for chairing the session earlier in the week. For those not present here are the [minutes](https://www.w3.org/2020/10/27-uach-minutes.html).

We (51Degrees) will update the issues early next week and provide an alternative proposal for the data structure in due course.

We felt the session was productive and would like to arrange another session to progress more of the issues with @travisleithead or another neutral chair.

miketaylr:
> We (51Degrees) will update the issues early next week and provide an alternative proposal for the data structure in due course.

My read on this comment (and the discussion) is that it's OK to close this PR now, is that correct @jwrosewell?

jwrosewell:
@miketaylr I would prefer to leave the pull request open as it provides specific text changes to the document. I have observed PING and TAG editors prefer specific text changes to open ended issues, presumably because it provides a concrete modification others can evaluate.

What are others views on arranging a follow on F2F session to progress the many issues during November?

miketaylr:
> @miketaylr I would prefer to leave the pull request open as it provides specific text changes to the document.

FWIW, I agree with @yoavweiss ([in comment above](https://github.com/WICG/ua-client-hints/pull/124#issuecomment-691803400)) that opening smaller PRs that correspond to your issues is a cleaner route to evaluating them, rather than attempt to address 8 open issues in this PR.

jwrosewell:
Splitting out the PR is time consuming for a small company with limited resources like mine. Next time I will do exactly that first time around and raise individual PRs for each change.

I'd prefer to leave this open until we've settled each of the related issues and raised PRs for each of them using the text of this PR as the starting point. We can focus on the substance of the issue in the meantime.

If this PR remaining open causes a problem I've not seen to editors and administrators I'll seek to find the time to split this out sooner.

jwrosewell:
@miketaylr I have now created the pull requests suggsted. Would you like to setup a group call to talk through each of the areas? I believe this would be more time efficient and productive than offline commenting.

miketaylr:
> I have now created the pull requests suggsted.

Thanks for the proposed contributions, @jwrosewell. I've commented on a few of them, and closed some others that should probably start as issues (especially where the PR just adds an `Issue: ` block to the spec text). Once you file those, let's continue discussions there. 

I think we can close this PR now, and perhaps refer back to some of the proposed language and points raised in the comments as needed.

PR #139, opened 2020-10-28
Add a considered alternatives section

### LINKED ISSUE #129: Alternatives require greater consideration
There is a problem acknowledged in the AC forum [https://lists.w3.org/Archives/Member/w3c-ac-forum/2020JulSep/0133.html] associated with the name incubation in the W3C Process. It implies the proposer desires the proposal to be successful and sends the wrong message to stakeholders unfamiliar with the W3C Process. The word “experiment” is often used as a synonym for “incubation”. This is a failing in the process which is better described by “experimentation”.  The process of experimentation yields information from which to then form a conclusion. It is open to the possibility of a failure path.

As an experimental proposal alternative solutions need to be considered and dismissed to justify why this experiment is needed over other alternatives. What work has been done in this area? How did those experiments progress?

One such alternative could include providing advice to implementors to gradually restructure the conventions associated with the existing User-Agent header field value to address one of the problems identified concerning ease of data consumption.

See [this pull request]( https://github.com/WICG/ua-client-hints/pull/124) which incorporates experiment feedback and review from interested users and other stakeholders for proposed text changes.

yoavweiss:
Closes #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,3 +570,16 @@ such, servers will usually use the platform or presence of this "mobile" identif
 pointing out that most modern browsers also have an explicit "request desktop site" UI element in their
 mobile versions which should be honored. In a more general sense, though, a "mobile" experience could be
 seen as a UX designed with smaller screens and touch-based interface in mind.
+
+# Considered alternatives
+## Freezing the UA string and reducing its information density without providing an alternative mechanism
+We've considered removing information from the UA string, similar to WebKit's [attempt](https://webkit.org/blog/8042/release-notes-for-safari-technology-preview-46/#post-8042:~:text=Froze%20the%20user%2Dagent%20string%20to%20reduce,to%20prevent%20its%20use%20for%20fingerprinting).
+
+We ruled out this alternative as it seems to leave many use-cases unaddressed.
+
+## Restructuring of the User-Agent string
+
+Restructuring the User-Agent string could have addressed some of the compatibility concerns with today's use of the User-Agent string, and potentially discourage its abuse.
+At the same time, it would have incurred same or higher migration costs as this proposal. In particular, attempts to restructure the string in-place would result in a lot of breakage of legacy apps.
+On top of that, it wouldn’t have addressed any of the passive entropy concerns that are motivating this proposal.
+


amtunlimited (approved):
I feel like we're missing something, but I can't remember...I'll look through my notes and make a new pull request if I find anything.

LGTM

Steve51D:
Hi @yoavweiss,
I feel like the restructuring of the User-Agent approach needs more consideration. Mostly because it seems that the `same or higher migration costs` statement is not really backed up by anything here. Is there additional work that could be referenced that would back this up?

For example, consider a simple case such as web request logging. Currently, the user-agent making the request is often in these logs as a single string. 
If the user-agent header is restructured then this doesn't need to change. However, if the client hints proposal moves forward as-is, then the logging code and any log management software will need to change to either mash all the client hints headers into a single string or support all the new headers.

My point is not that one way is good and another is bad, but that I'm sure there are some areas that are more affected by restructuring the user-agent and others that are more affected by switching to something like client hints. I think it's worth understanding what those areas are before discarding one approach for the other.

miketaylr:
I'm inclined to merge this as-is. 

I can appreciate @Steve51D's comments though -- if there are concrete improvements you'd like to suggest, we can consider those in a future PR or new issue. Thanks!

(it's true that for logs, getting a (new, restructured) string to a file has a zero-migration cost, but tooling that parsed said logs may or may not have to be updated. but we can discuss in a new issue. ^__^)

jwrosewell:
In relation to the now merged issue and question raised. 

I was pointing to my position in relation to a debate within the W3C AC concerning this matter of which this was a summary post.

Since posting the authoritative voice of [US Congress](https://github.com/WICG/ua-client-hints/files/5482597/US.Congressional.Report.competition_in_digital_markets.pdf) has raised the following observations in relation to this general issue on page 229.

> First, changes to Chrome’s functionality create de facto standards. Market participants must adhere to these standards or risk their technology no longer being compatible with most websites. Market participants explain that Google will often build features quickly without using the standard setting process or giving smaller browsers time to implement new features. Once web developers start building to these specifications, however, smaller browsers are under pressure to quickly implement these changes, often with little notice. If smaller browsers cannot keep up, users are flooded with “[b]rowser not supported” messages on webpages that have already built to Chrome’s specifications. Several market participants told the Subcommittee that they felt “bullied” by this process.   

> Second, Google’s has an outsized role in the formal stakeholder standards-making processes. As explained earlier in this Report, the World Wide Web Consortium (W3C) is one of the leading standards organizations in the browser market. Its stated mission is to be “open and collectively empowering.” Other market participants believe that Google is significantly overrepresented in the W3C web platform incubator community group (WICG). They note that Google’s employees comprise 106 members, more than eight times the number of employees from Microsoft, the next largest stakeholder represented. Most companies, meanwhile, have only one representative. One market participant said:  
> "Though standards bodies like the W3C give the impression of being a place where browser vendors collaborate to improve the web platform; in reality Google’s monopoly position and aggressive rate of shipping non-standard features frequently reduce standards bodies to codifying web features and decisions Google has already made"

I personally recognize the statements made and believe the first observation would be correctly applied not just to small browser vendors but other stakeholders in the web eco system.

It is beholden on Google as a dominant market player to present alternative solutions, debate these with impacted stakeholders in a fair and level forum free from intimidation, and then gain consensus before presenting a single solution to the wider world. To do otherwise is to present the impression of a "done deal".

Whilst the alternatives section has been expanded it does not currently give serious consideraitons to all the options, how they would address the problems, or the mertis or weaknesses of each compared to others. Without such analysis it is impossibe to explain why one solution was favored over another.

cwilso:
James, the above comment by @miketaylr explicitly invites future pull requests of additional language to improve this section, or additional issue filing to discuss further considerations.  I believe this pull request (which is now merged) is a substantial step in the right direction, but you are explicitly invited to help improve the effort.

(As stated separately, it is a generally preferred practice to file concise, single-issue pull requests, as they are easier to reach consensus on and merge.)

PR #154, opened 2020-11-02
Fixes #141 - Use allowed values in platform architecture examples

### LINKED ISSUE #141: Examples for "platform architecture" don't match the actually allowed values
The [section that defines client hints](https://wicg.github.io/ua-client-hints/#http-ua-hints) says:

> platform architecture - The user agent's underlying CPU architecture (for example: "ARM64", or "ia32")

But below:

> User agents MUST map higher-entropy platform architecture values to the following buckets:
> x86 CPU architectures => "x86"
> ARM CPU architectures => "arm"

The example should probably match the allowed values, instead of using different names for those same architectures.

miketaylr:


miketaylr:
> miketaylr did not make IPR commitments for this group. To make the IPR commitments, miketaylr should join the Web Platform Incubator Community Group .

oops. brb.

amtunlimited (approved):

PR #158, opened 2020-11-04
Make clear in `create brand` section that randomization can happen at build time

### LINKED ISSUE #143: Non-normative Note on GREASE may be inconsistent with the normative algorithm
The [create brands algorithm](https://wicg.github.io/ua-client-hints/#create-ua-list-section) includes the step:

> Randomize the order of the items in list.

In [a different section](https://wicg.github.io/ua-client-hints/#grease), a non-normative note says:
> Note: One approach to minimize caching variance could be to determine the GREASE parts of the UA set at build time, and keep them identical throughout the lifetime of the user agent's significant version.

On the face of it, this seems inconsistent, since the algorithm says to randomize, and it's normally the case that spec algorithms run at runtime, not compile time. Reading more closely, it is not actually specified when or how often **create brands** runs, only that [**brands** is generated from it](https://wicg.github.io/ua-client-hints/#monkeypatch-html-windoworworkerglobalscope).

If it's actually intended that the spec allows the unusual behavior of running an algorithm at UA compile time, then it would probably be best for the Note to be right next to the definition of **brands** or **create brands**. If not, then the note should be revised.

amtunlimited:
@miketaylr @othermaciej

***
<a href="https://pr-preview.s3.amazonaws.com/amtunlimited/ua-client-hints/pull/158.html" title="Last updated on Nov 4, 2020, 9:27 PM UTC (bb0ade7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/158/dbb11ce...amtunlimited:bb0ade7.html" title="Last updated on Nov 4, 2020, 9:27 PM UTC (bb0ade7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -430,7 +430,11 @@ When asked to run the <dfn>create brands</dfn> algorithm, the [=user agent=] MUS
 
     2.  Randomize the order of the items in |list|.
 
-    Note: See [[#grease]] for more details on why these steps might be appropriate.
+    Note: One approach to minimize caching variance when generating these random components could be to
+    determine them at build time, and keep them identical throughout the lifetime of the [=user agent=]'s significant
+    version. 
+
+    Note: See [[#grease]] for more details on when and why these randomization steps might be appropriate.
 
 5. Return |list|.
 


yoavweiss (approved):
LGTM

PR #159, opened 2020-11-05
Fix #150. Make JS interface in explainer question more specific

### LINKED ISSUE #150: Explainer question about neutering the JavaScript interface has an unclear referent
https://github.com/WICG/ua-client-hints#do-we-really-need-to-neuter-the-javascript-interface-too

> Do we really need to neuter the JavaScript interface too?

What is this referring to? The explainer (like the spec) describes a JavaScript interface that has access to all the same info as can be retrieved via HTTP, so in what way is it neutered? This question should probably explain what it's talking about.

amtunlimited:


miketaylr (approved):
LGTM, thanks

Issue #151, opened 2020-11-02
getHighEntropyValues should not return all values, only values that have been opted in


othermaciej:
The definition of [`getHighEntropyValues`](https://wicg.github.io/ua-client-hints/#getHighEntropyValues) sounds like it might return all the high entropy values, not just ones that have been opted in. That seems inconsistent with the rest of the spec, which requires prior opt-in. Which is intended?

If this exposes all values, including ones that have not been opted in, then I have a strong objection to this requirement, as it would allow third party scripts embedded in the first party context to request all entropy values, which dramatically increases fingerprinting surface that can be exposed without active cooperation of the first party.

A non-normative "Note" says that user-agents may "gate their exposure behind potentially time-consuming checks", but it doesn't give a means to answer some, but not all of the questions, nor is this possibility expressed in the definition of the algorithm for `getHighEntropyValues`, which is written as if the promise always resolves successfully and always returns all values.

I would strongly recommend returning only the values that the server has opted into via headers, and also explicitly giving license to UAs to reject the promise, or to fail to fill in some of the values.

Another potential problem: it seems like as written, this measure should work even in third-party iframes that the top level document has not delegated permission to. The Client Hints infrastructure covers delegation and permissions for the HTTP headers, but not, I think, for this JavaScript interface.

jonarnes:
Related to #37 The js API does not require opt-in.

othermaciej:
Updating title based on the clarification from @jonarnes . With that clarification this is a critical issue for Safari; we would not implement UA client hints unless this is addressed. (We otherwise feel most of our critical issues have been resolved).

yoavweiss:
Bringing back the discussion from [webkit-dev](https://lists.webkit.org/pipermail/webkit-dev/2020-November/031576.html) (my bad!) to here.

I believe that having per-hint Permission Policy to enable the high entropy bits of the JS API (similar to the [policies](https://wicg.github.io/client-hints-infrastructure/#policy-controlled-features) we use for delegation) will give us the ability to both enable the JS API on specific iframes, or disable it entirely (even for the main frame and any scripts that run in its context).

@clelland to keep me honest.

I prefer to keep the Client Hints opt-in as something that's responsible for the request headers but not more than that, as it would start overlapping with Permission Policy.

 > The spec does not eve allow denying the request entirely as written. A non-normative Note suggests that is allowed, but I can’t find any step in the algorithm that would ever reject the promise.

That sounds like a bug that we should fix.

[...]

PR #163, opened 2020-12-16
Issue #151 - Make it possible to reject a promise from getHighEntropyValues

miketaylr:
This partially addresses #151 by allowing UAs to reject the promise created from getHighEntropyValues. Once we have consensus on the rest, we can remove the issue and clarify what it means to "decide".

PTAL @amtunlimited

cc @othermaciej

***
<a href="https://pr-preview.s3.amazonaws.com/miketaylr/ua-client-hints/pull/163.html" title="Last updated on Dec 16, 2020, 5:19 PM UTC (11548b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/163/8ee4664...miketaylr:11548b9.html" title="Last updated on Dec 16, 2020, 5:19 PM UTC (11548b9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -441,7 +441,7 @@ When asked to run the <dfn>create brands</dfn> algorithm, the [=user agent=] MUS
 
     Note: One approach to minimize caching variance when generating these random components could be to
     determine them at build time, and keep them identical throughout the lifetime of the [=user agent=]'s significant
-    version. 
+    version.
 
     Note: See [[#grease]] for more details on when and why these randomization steps might be appropriate.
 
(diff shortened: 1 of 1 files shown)


miketaylr:
*gentle ping* @amtunlimited ❤️ 🌈 🐰

amtunlimited:
Sorry about the delay. Been a crazy...time?

Quick question: If I'm reading this correctly, does this mean that if one hint is rejected, the client does not get any hints?

miketaylr:
> Quick question: If I'm reading this correctly, does this mean that if one hint is rejected, the client does not get any hints?

Correct. You don't get *any* hints from `getHighEntropyValues` if the UA decides you can't have one. Right now there's no platform-notion of "partially resolve a promise", and it seems better to know up front that you didn't get some of the data you wanted then have to double check every item you did get back for a non-null result.

Perhaps a non-normative Note would be useful to encourage UAs to give some more context as to why they just sent a `NotAllowedError`?

amtunlimited:
> Perhaps a non-normative Note would be useful to encourage UAs to give some more context as to why they just sent a NotAllowedError?

IIRC we explicitly didn't do this for the headers (e.g. setting some sort of "not set" header). If I can find the relevant discussion I will link it here.

Is there a notion of "optional" in JS to represent not-present vs intentionally blank?

miketaylr:
> Is there a notion of "optional" in JS to represent not-present vs intentionally blank?

Yeah, that's what `null` represents (vs `undefined`). I thought about returning `null` for values that the UA wants to refuse, but I think getting a rejected promise for the whole lot is conceptually simpler.

amtunlimited (approved):
LGTM

Issue #155, opened 2020-11-03
Impact on same-origin policy


annevk:
These new headers increase the size of an HTTP request and coupled with attacker-controlled headers or header values could be used to carry out certain cookie-size sniffing attacks.

Privacy measures in browsers might invalidate some of these attacks, but the privacy boundary is typically not the origin, at least in today's implementations.

miketaylr:
Attaching some additional context, indirectly via @annevk:
*  https://bugzilla.mozilla.org/show_bug.cgi?id=1447631
* https://github.com/whatwg/fetch/issues/903
* https://github.com/whatwg/fetch/pull/736

yoavweiss:
Looking at table 2 in the paper quoted in that Bugzilla issue, it seems like the underlying concern is that we need to make sure that *overall* request header sizes should be below 8KB in order for them to be CORS safe.
The specific vector described to distinguish 200 from a 400 was since fixed, but maybe there are others.

This issue seems to be impacting all new request headers, not just UA-CH, or even CH in general. e.g. `Sec-Fetch-*`
/cc @mikewest 

We cannot place a CORS cap on the overall request headers size, as this would create its own side channel. At the same time, we can (and [looks like we have](https://fetch.spec.whatwg.org/#cors-unsafe-request-header-names)) add a cap on the overall size of safe-listed headers, as 1024.

So, if I'm reading this correctly, UA-CH and other safelisted headers don't impact same-origin policy beyond running a risk of triggering a preflight, if the list gets too long. Is that correct?

annevk:
Well, it's unclear, because it's not defined how `UA-CH` et al work.

yoavweiss:
> We cannot place a CORS cap on the overall request headers size, as this would create its own side channel. At the same time, we can (and [looks like we have](https://fetch.spec.whatwg.org/#cors-unsafe-request-header-names)) add a cap on the overall size of safe-listed headers, as 1024.

@annevk - reviving this part. Would it make sense to add a cap on the size of safelisted headers added per request? such that if someone is adding *all* the Client Hints, they run a risk of triggering preflights, but that won't be an issue in the typical case?

annevk:
I think coupled with `Sec-*` that would indeed address the concern. It's a bit unclear to me how that would work in practice. And there's a problem with navigations as those don't do preflights at the moment.

yoavweiss:
Should we move this to client-hints-infra? This doesn't seem specific to UA-CH

[...]

Issue #152, opened 2020-11-02
3rd-party access to client-hints causes payload bloat


Steve51D:
Currently, 3rd parties will only be sent client hints if both:

- Accept-CH is set to request the relevant headers.
- Feature-Policy / Permissions-Policy is set to allow each 3rd party access to each specific header.

For major websites with a large web of 3rd party dependencies, this ends up being incredibly verbose and can significantly inflate the size of the response.

Is there any reason that a more succinct solution, even a re-use of existing headers such as content-security-policy, is not viable?
For example, if I've given permissions for JavaScript from www.example.com to run on my page using csp, why should I then need to separately allow access to client hints for that 3rd party?

Steve51D:
Looking through the permissions-policy issues list, I think this may already be covered by w3c/webappsec-permissions-policy#408

miketaylr:
> For major websites with a large web of 3rd party dependencies, this ends up being incredibly verbose and can significantly inflate the size of the response.

It would be interesting to see some data on this. Do you have any examples with measurements @Steve51D?

eeeps:
There's also https://github.com/WICG/client-hints-infrastructure/issues/23

My complaint there stems from how awkward this is to explain to people. I've settled on:

"Accept-CH asks the browser to send particular hints. Permissions-Policy tells the browser to send particular hints to particular origins. It seems a bit duplicative because it is: you need to do both."

PR #165, opened 2020-12-18
Amend the chapter on delegation to make this simpler and respect trust signal already provided.

jwrosewell:
Web site operators should be free to choose their suppliers without undue interference or complexity. This modification respects that right and simplifies the proposal. Relates to issue [152](https://github.com/WICG/ua-client-hints/issues/152) and [155](https://github.com/WICG/ua-client-hints/issues/155).

***
<a href="https://pr-preview.s3.amazonaws.com/jwrosewell/ua-client-hints/pull/165.html" title="Last updated on Mar 5, 2021, 9:43 PM UTC (4062560)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/165/0edcf9d...jwrosewell:4062560.html" title="Last updated on Mar 5, 2021, 9:43 PM UTC (4062560)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -491,11 +491,16 @@ of a given agent's behavior over time.
 Delegation {#delegation}
 ----------
 
-Client Hints will be delegated from top-level pages via Feature Policy. This reduces the likelihood that [=user agent=]
-information will be delivered along with subresource requests, which reduces the potential for
-passive fingerprinting.
+All requests initiated by the user agent (aka as third party) will contain the
+same Client Hint headers as those requested by the origin (also know as first 
+party). For example if the a request to https://example.com included the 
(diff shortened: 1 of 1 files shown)


miketaylr:
The current model of requiring delegation to third parties does not prevent first parties from trusting (and hence delegating to) third parties, it just requires it to be explicit. This isn't something we would like to change now, but thanks for the PR.

jwrosewell:
The first party would not include the third party in the page if they did not want the third party to be there. It is already explicit. Adding multiple layers and complexity adds nothing except complexity and associated cost for those migrating. It also adds latency which is particularly problematic for smaller web participants.

PR #166, opened 2020-12-18
Removed duplication associated with secure transports

jwrosewell:
This is already covered in the dependent specification concerning client hints.

***
<a href="https://pr-preview.s3.amazonaws.com/jwrosewell/ua-client-hints/pull/166.html" title="Last updated on Dec 18, 2020, 5:17 PM UTC (cca3a5e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/166/0edcf9d...jwrosewell:cca3a5e.html" title="Last updated on Dec 18, 2020, 5:17 PM UTC (cca3a5e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -480,14 +480,6 @@ The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></df
 Security and Privacy Considerations {#security-privacy}
 ===================================
 
-Secure Transport {#secure-transport}
-----------------
-
-Client Hints will not be delivered to non-secure endpoints (see the secure transport requirements in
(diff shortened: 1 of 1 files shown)


miketaylr:
Hi @jwrosewell, thanks for the PR. However, I happen to think having possibly redundant information related to security (or privacy) is a win, especially for folks who read this document before following any possible outbound links. As such, I think we should leave it in place. If you want to discuss further, let's do it in https://github.com/WICG/ua-client-hints/issues/181 -- which should tackle the dependencies question.

PR #167, opened 2020-12-18
Added section to clearly explain what is out of scope in the proposal.

jwrosewell:
A technical specification is only concerned with the problem being solved and the technical design. Implementation is a different subject.


miketaylr:
Thanks for the PR, but please open an issue for discussion first, per https://github.com/WICG/ua-client-hints/blob/main/CONTRIBUTING.md#pull-requests (that note was added after this PR, but it's not entirely clear to me what this PR is intending to accomplish).

Issue #134, opened 2020-09-23
Default policy


jwrosewell:
What is the policy for classifying some Client Hints as being sent by default? Either all Client Hints should be sent on every request, or no Client Hints sent on first request. If the rationale for the specification is to limit entropy then sending a value will increase entropy for minority browser vendors, whilst decreasing it for dominate browser vendors. This places minority browser vendors at a disadvantage. If the rationale is to address historic issues with the `User-Agent` field value construction this could be better addressed via a simpler specification to introduce a more modern convention for the field.

See [this pull request]( https://github.com/WICG/ua-client-hints/pull/124) which incorporates experiment feedback and review from interested users and other stakeholders for proposed text changes.

jwrosewell:
I suppose this could be a duplicate concerning the definition of high low entropy IF the concept of restricting information is not dependent on these terms in any way. If default are dependent on high and low entropy then they will need to be defined a policy for catagorising them defined.

yoavweiss:
The rationale of sending some hints by default (brand, version and mobileness) was to significantly reduce the entropy exposed by user agent information while still addressing many of the [use cases](https://github.com/WICG/ua-client-hints#use-cases) by default.

The use cases that were deemed critical to enable by default:
* Differential serving - community feedback we got was that this is a [critical use-case](https://groups.google.com/a/chromium.org/g/blink-dev/c/A4wxFpvqUfA/m/bzCc9XW0CgAJ) for both [polyfill serving](https://groups.google.com/a/chromium.org/d/msgid/blink-dev/d5087dcc-aa14-4907-85c2-085a5ca3b5b5%40chromium.org) and bug workarounds.
* Analytics - folks working on less-tested browsers expressed concerns that without accurate analytics, sites will not care about testing in their browser, resulting in compatibility issues. Therefore, it seems like a major use-case to enable.
* Mobile site content adaptation - [community feedback](https://github.com/WICG/ua-client-hints/issues/15) indicated that this use case is a critical one, and exposing the extra entropy to enable it is worthwhile.

As a [response](https://groups.google.com/a/chromium.org/g/blink-dev/c/A4wxFpvqUfA/m/yBjL6B0QDgAJ) to said community feedback, we adjusted the exposed-by-default hints of this proposal.

Note that browsers are still free to e.g. mimic other browsers when sending those by-default hints, in case they care more about exposed entropy than the above use-cases.

miketaylr:
[...]

Issue #131, opened 2020-09-23
Determination of high and low entropy Client Hints


jwrosewell:
How are low and high entropy Client Hints determined in the Client Hints Infrastructure document? How are they balanced against considerations other than privacy? For example, a company with access to the equivalent data via other means such as a header sent only to their services might seek to restrict access to some values citing entropy as justification. This needs to be resolved before referencing in this document.

See [this pull request]( https://github.com/WICG/ua-client-hints/pull/124) which incorporates experiment feedback and review from interested users and other stakeholders for proposed text changes.

yoavweiss:
This seems like a duplicate of #134

PR #173, opened 2020-12-18
Added issue related to the definition of high or low entropy.

jwrosewell:
Need definition for implementors concerning how classifications are determined. Guidance is needed to ensure information asymmetries do not exist between user agent vendors and the wider web eco system.

Relates to issue [134](https://github.com/WICG/ua-client-hints/issues/134) and [131](https://github.com/WICG/ua-client-hints/issues/131).

***
<a href="https://pr-preview.s3.amazonaws.com/jwrosewell/ua-client-hints/pull/173.html" title="Last updated on Dec 18, 2020, 5:35 PM UTC (d27673e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/173/0edcf9d...jwrosewell:d27673e.html" title="Last updated on Dec 18, 2020, 5:35 PM UTC (d27673e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -314,6 +314,12 @@ and the significant version number (both of which are fairly clearly sniffable b
 structure of other headers and by testing for the availability and semantics of the features
 introduced or modified between releases of a particular browser" [[Janc2014]]).
 
+ISSUE: What is the policy for classifying some Client Hints as being sent by default? Either all Client 
+Hints should be sent on every request, or no Client Hints sent on first request. If the rationale for
+the specification is to limit entropy then sending a value will increase entropy for minority browser 
+vendors, whilst decrease it for dominate browser vendors. If the rationale is to address historic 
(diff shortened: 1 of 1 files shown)


miketaylr:
Hi @jwrosewell, rather than add an Issue block to the spec, can you please open a new issue on the repo? I would prefer we align on an answer to the question in an issue and fix it by adding the resolution directly into the spec, rather than encode unresolved issues into the spec (where possible). I think we can clarify this particular issue around high vs low entropy and default hints.

Thanks!

(Also note, that IPR needs to be handled, see https://labs.w3.org/repo-manager/pr/id/WICG/ua-client-hints/173)

miketaylr:
> can you please open a new issue on the repo?

Apparently I didn't have much caffeine when I looked at this. Issue #134 exists, so we can continue conversation there.

Issue #161, opened 2020-11-18
Escaped string and backslash not web compatible


miketaylr:
Based on Chrome's experimental roll-out of `Sec-UA-CH` we've concluded that `\\` and `\"` in the arbitrary brand value is not web compatible.

The biggest hurdle (we're aware of) is the ESAPI library, which would [require a re-write](https://github.com/ESAPI/esapi-java-legacy/issues/578#issuecomment-725710383) to handle these escaped chars correctly.

* [1145279](https://bugs.chromium.org/p/chromium/issues/detail?id=1145279): 404 on soundtransit.org
  * This is caused by the \\ and \”, removing them fixes the issue.
  * Worth noting “X-Generator: Drupal 8 (https://www.drupal.org)” in the 200 response -- they’re probably not using ESAPI (unless they have a hybrid PhP / Java stack)
* ESAPI rejects requests due to \\ and \”
  * https://bugs.chromium.org/p/chromium/issues/detail?id=1142532
  * https://bugs.chromium.org/p/chromium/issues/detail?id=1140849 
* ESAPI multiple encoding issue (filed mentioning Chrome explicitly)
  * https://github.com/ESAPI/esapi-java-legacy/issues/574 
* [1144295](https://bugs.chromium.org/p/chromium/issues/detail?id=1144295): LiveSite blocking UA-CH as multiple encoding breaking www.leedsbuildingsociety.co.uk and probably more sites. 
* [1147471](https://bugs.chromium.org/p/chromium/issues/detail?id=1147471): Unable to get business insurance quote on https://smallbizquote.thehartford.com/
  * The server is rejecting any quote in a header value
  * They’re also [using ESAPI](https://bugs.chromium.org/p/chromium/issues/detail?id=1147471#c16), it turns out

For now, we're unshipping the escaped chars from Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1149575

The spec should be updated to reflect what's safe to ship on the web (without breaking it). 

Note: This will obsolete https://github.com/WICG/ua-client-hints/issues/114

Issue #146, opened 2020-11-01
GREASE algorithm is underspecified


othermaciej:
The [**create brands** algorithm](https://wicg.github.io/ua-client-hints/#create-ua-list-section) requires adding arbitrary values, but does not define what these should be, how they should be generated, or how often they should change. If UAs use very different arbitrary values, it's possible sites will intentionally or inadvertently start to depend on a particular value or set of value. Since UAs may need to emulate the values of other UAs for compatibility reasons, this may end up requiring reverse engineering of the arbitrary values.

miketaylr:
> If UAs use very different arbitrary values, it's possible sites will intentionally or inadvertently start to depend on a particular value or set of value.

Good point, e.g., if Safari always uses 3 semicolons in its grease values, but Chrome always uses 2, at some point developers will figure that out and fork on behavior (if history teaches us anything).

> Since UAs may need to emulate the values of other UAs for compatibility reasons, this may end up requiring reverse engineering of the arbitrary values.

I think the GREASE model will have failed if UAs have to emulate not only "real" branding and version inside the UA-CH list, but also the arbitrary values. But I also agree that coming up with these values could be more concrete to avoid this situation.

othermaciej:
> I think the GREASE model will have failed if UAs have to emulate not only "real" branding and version inside the UA-CH list, but also the arbitrary values.

I agree that GREASE would have failed in this circumstance; this proposal is intended to reduce the odds that it would fail in this way.

miketaylr:
Hey @othermaciej, I've rewritten the GREASE algorithms and would appreciate any feedback you have:

See https://wicg.github.io/ua-client-hints/#create-ua-list-section and 
https://wicg.github.io/ua-client-hints/#create-arbitrary-brands-section, in particular. Thanks.

othermaciej:
Hmm, no, I think those changes still leave the algorithm under specified in exactly the way stated in the original issue. I said:

> The [create brands algorithm](https://wicg.github.io/ua-client-hints/#create-ua-list-section) requires adding arbitrary values, but does not define what these should be, how they should be generated, or how often they should change.

https://wicg.github.io/ua-client-hints/#create-an-arbitrary-brand starts with this step:
[...]

PR #178, opened 2021-01-05
Issue #146 - Specify GREASE algorithms (aka arbitrary brand and version number)

miketaylr:
PTAL @amtunlimited 

(Right now I'm running a crawler to ensure `_ ( ) ; ? = - : . /` are safe to use, but so far it's looking good).

This also should address https://github.com/WICG/ua-client-hints/issues/161.

***
<a href="https://pr-preview.s3.amazonaws.com/miketaylr/ua-client-hints/pull/178.html" title="Last updated on Jan 7, 2021, 5:35 PM UTC (f9e8cf5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/178/85c1b52...miketaylr:f9e8cf5.html" title="Last updated on Jan 7, 2021, 5:35 PM UTC (f9e8cf5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -422,8 +422,8 @@ Every {{WindowOrWorkerGlobalScope}} object has an associated <dfn for="WindowOrW
 When asked to run the <dfn>create brands</dfn> algorithm, the [=user agent=] MUST run the following steps:
 1. Let |list| be a [=/list=].
 
-2. Collect pairs of [=user agent/brand=] and [=user agent/significant version=] which represent the [=user agent=],
-    its equivalence class and/or its rendering engine.
+2. Collect pairs of [=user agent/brand=] and [=user agent/significant version=] which represent the [=user agent=] or
+    [=equivalence classes=].
 
 3. For each pair:
 
@@ -441,11 +441,14 @@ When asked to run the <dfn>create brands</dfn> algorithm, the [=user agent=] MUS
 
     Note: One approach to minimize caching variance when generating these random components could be to
     determine them at build time, and keep them identical throughout the lifetime of the [=user agent=]'s significant
-    version. 
+    version.
 
     Note: See [[#grease]] for more details on when and why these randomization steps might be appropriate.
 
 5. Return |list|.
+
+An <dfn for="user agent" export>equivalence class</dfn> represents a group of browsers believed to be compatibile with each other. A shared
+    rendering engine may form an [=equivalence class=], for example.
 
 <h3 id="getters">Getters</h3>
 
@@ -548,10 +551,10 @@ Let's examine a few examples:
 * In order to avoid sites from barring unknown browsers from their allow lists, Chrome could send a
     UA set that includes an non-existent browser, and which varies once in a while.
     - `"Chrome"; v="73", "NotBrowser"; v="12"`
-* In order to enable equivalence classes based on Chromium versions, Chrome could add the rendering
+* In order to enable [=equivalence classes=] based on Chromium versions, Chrome could add the rendering
     engine and its version to that.
     - `"Chrome"; v="73", "NotBrowser"; v="12", "Chromium"; v="73"`
-* In order to encourage sites to rely on equivalence classes based on Chromium versions rather than
+* In order to encourage sites to rely on [=equivalence classes=] based on Chromium versions rather than
     exact UA sniffing, Chrome might remove itself from the set entirely.
     - `"Chromium"; v="73", "NotBrowser"; v="12"`
 * Browsers based on Chromium may use a similar UA string, but use their own brand as part of the
--- a/index.bs
+++ b/index.bs
@@ -602,11 +602,6 @@ Let's examine a few examples:
 [=User agents=] MUST include more than a single value in [=user agent/brands=], where one of these values is an
 arbitrary value.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -411,49 +411,89 @@ Note: The high-entropy portions of the user agent information are retrieved thro
...
 3. For each pair:
 
     1. Let |dict| be a new {{NavigatorUABrandVersion}} dictionary,
-        with [=user agent/brand=] as {{NavigatorUABrandVersion/brand}} and [=user agent/significant version=] as {{NavigatorUABrandVersion/version}}.
+        with [=user agent/brand=] as {{NavigatorUABrandVersion/brand}} and [=user agent/significant version=] as
+        {{NavigatorUABrandVersion/version}}.
 
     2. Append |dict| to |list|.
 
 4.  The [=user agent=] SHOULD execute the following steps:
```
amtunlimited, line 439: Somewhat unfamiliar with spec language; does this wording imply that if you don't do one, you shouldn't do the other?

index.bs
```diff
@@ -411,49 +411,89 @@ Note: The high-entropy portions of the user agent information are retrieved thro
...
-<h3 id="getters">Getters</h3>
+An <dfn for="user agent" export>equivalence class</dfn> represents a group of browsers believed to be compatibile with
+each other. A shared rendering engine may form an [=equivalence class=], for example.
+
+<h4 id="create-arbitrary-brands-section"
+    algorithm="to create arbitrary brand and version values">Create arbitrary brand and version values</h4>
+
+To <dfn>create an arbitrary brand</dfn>, the [=user agent=] MUST run these steps:
+
+    1. Let |arbitraryBrand| be a [=/string=] composed of [=ASCII lower alpha=], [=ASCII upper alpha=], or 0x20 (SP), no
```
amtunlimited, line 463: Is the char-set or byte limit imposed on normal brands as well? Should it be?

index.bs
```diff
@@ -411,49 +411,89 @@ Note: The high-entropy portions of the user agent information are retrieved thro
...
+         longer than twenty [=ASCII bytes=].
+    1. Let |arbitraryBrandList| be the result of <a lt="split on ASCII whitespace">splitting |arbitraryBrand| on ASCII
```
amtunlimited, line 465: Should there be a minimum number of spaces included to ensure there are a minimum number of GREASEy chars?

index.bs
amtunlimited, line 463: Should we include numbers in this range?

amtunlimited, line 463: (hat trick!) nit: s/`[=ASCII lower alpha=], [=ASCII upper alpha=]`/`[=ASCII alpha=]`/ (or [=ASCII alphanumeric=] if we decide to add numbers into the mix)

index.bs
```diff
@@ -411,49 +411,89 @@ Note: The high-entropy portions of the user agent information are retrieved thro
...
+    1. Let |greaseyChars| be the [=list=] of [=ASCII bytes=] « 0x20 (SP), 0x28 (left parenthesis), 0x29
+         (right parenthesis), 0x2D (-), 0x2E (.), 0x2F (/), 0x3A (:), 0x3B (;), 0x3D (=), 0x3F (?), 0x5F (_) ».
+    1. For each item of |arbitraryBrandList|, [=stack/push=] a randomly selected [=list/item=] from |greaseyChars| onto
+         |greaseyStack|.
+    1. Let |greaseyBrandList| be a [=list=] and |index| be 0.
+    1. While |greaseyStack| [=stack/is not empty=]:
+        1. Let |item| be the result of <a lt="pop">popping</a> from |greaseyStack|.
+        1. Append |item| to |greaseyBrandList|.
+        1. Append |arbitraryBrandList|[|index|] to |greaseyBrandList|.
+        1. Increment |index| by 1.
```
amtunlimited, line 477: Might this end up in an array-out-of-bounds?

index.bs
miketaylr: Yes, if you were to only pick one, you would be violating the spec.

index.bs
miketaylr: It's not, but we can open an issue to consider it: https://github.com/WICG/ua-client-hints/issues/179

index.bs
miketaylr: That's an interesting point. If we require at least 1 space, you end up with at least 2 greasy chars. WDYT @amtunlimited?

index.bs
miketaylr: I think we shouldn't. I don't think there's a good reason to encourage parsing of numbers in a brand name, instead consumers should look in the version field.

(And I double checked that there's nothing surprising here:
https://en.wikipedia.org/wiki/List_of_web_browsers)

miketaylr: Oh nice, thanks!

index.bs
miketaylr: (I think I'm going to make 1 space a requirement, but open to other ideas)

index.bs
miketaylr: I don't see how, do you have a scenario where it might?

amtunlimited: I misread step 5; silly me

amtunlimited (approved):
LGTM

miketaylr:
Thanks for the review!

PR #197, opened 2021-02-09
Analytics ecological pitfalls; recc. treating brands list as unit

### LINKED ISSUE #115: Usage share analytics + GREASE + white lies for compatibility
The GREASE in the brands list effectively means that, looking at individual brands inside it, a browser recogniser must fail open and ignore anything it doesn't understand.

In the context of actually delivering a page and making it work, this is desirable: it promotes resillience and ought to gracefully degrade.

However, it's more troublesome when trying to use `Sec-CH-UA` for counting browser usage shares.

Consider a minority browser based on Chromium. Because it's a minority but likely quite Chrome-compatible, it's very possible that it'll claim to be Chrome in `Sec-CH-UA`, as well as its true name, relying on the GREASEing to keep the true name's presence from causing trouble as far as displaying pages works. So far, so good. Once the minority browser starts showing up in analytics, Web developers will start acknowledging its existence, and gradually the need to lie will fade away, right?

Except, how should usage share analytics work here? How can the minority browser's true name be distinguished from Chrome's GREASE? Because unknown brand names have to be ignored, the minority browser just gets bucketted as 'Chrome', meaning that the compatibility-versus-truthfulness dilemma persists - it doesn't show up in analytics, therefore no-one bothers to add its brand name to their recognisers.

The effect here will likely be that Chrome winds up with an inflated market share, only reinforcing its hegemony, and minority browsers have another impediment put in front of them. This is less than ideal for the health of the Web.

One thing to do here would be to recognise that analytics and page-display need different things from a browser recogniser. In the latter, just finding a brand that you recognise in the list + GREASE enforcing failing open is plenty good enough; failing open is definitely the thing to do, because graceful degradation. In the former, you do want to identify the exact browser and fail closed if you can't; analytics breaking doesn't break the page, so this is OK. How can this be made to work with GREASE?

The considerations around caching already mean that the spec recommends fixing the GREASE per browser version, and I think that that's probably enough to work with here: analytics could look at the set of brands & versions and match it as a whole against a known list sent by each browser version. It'd mean that the recogniser used for analytics would need to be updated for every new browser version; but, as noted, analytics breaking doesn't break the site, so this is palatable.

So, my concrete suggestion here is that no major change is needed here, but section 5.2 ('GREASE-like UA strings') could do with mentioning analytics as another reason to minimise within-version variance, parallel to the `SHOULD` about caching, and then some discussion in the explainer about analytics wanting to fail closed and looking at `Sec-CH-UA` as a unit while warning about the ecological pitfalls of trying to inspect it too deeply.

quasicomputational:
Closes #115.

This presumes that there's the suggested tweak to the spec as mentioned in #115.

cc @miketaylr

***
<a href="https://pr-preview.s3.amazonaws.com/quasicomputational/ua-client-hints/pull/197.html" title="Last updated on Apr 5, 2021, 3:42 PM UTC (224e1eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/197/2b830e6...quasicomputational:224e1eb.html" title="Last updated on Apr 5, 2021, 3:42 PM UTC (224e1eb)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ By design, looking at individual entries in the brands list makes it
 hard to distinguish between a less-popular browser's truthful brand
 name and a more-popular browser's arbitrary GREASE. Since the
 less-popular browser may include several popular brand names for
-compatibility purposes, its users will likely be bucketted as using
+compatibility purposes, its users will likely be bucketed as using
 the more-popular one if this approach is taken, leading to distorted
 views of usage share that favour already-popular browsers and with
 less-popular browsers possibly never gaining any visibility.
@@ -318,7 +318,7 @@ as a unit, and compare it to known lists of brands sent by various
 the various (browser, version) pairs that are to be distinguished.
 This will necessitate regular updates to the list of known lists of
 brands when new browser versions are released or new browsers become
-popular, or else everything will get bucketted as an unknown browser.
+popular, or else everything will get bucketed as an unknown browser.
 However, as this doesn't break sites for users, failing closed for
 unknown browsers is acceptable in this context.
 
--- a/README.md
+++ b/README.md
@@ -304,6 +304,32 @@ Sites that wish to provide market share analytics using UA-CH will need to
 inspect the `Sec-CH-UA` header, that is sent by default on every request, and
 keep a record of it.
 
+By design, looking at individual entries in the brands list makes it
+hard to distinguish between a less-popular browser's truthful brand
+name and a more-popular browser's arbitrary GREASE. Since the
+less-popular browser may include several popular brand names for
+compatibility purposes, its users will likely be bucketted as using
+the more-popular one if this approach is taken, leading to distorted
+views of usage share that favour already-popular browsers and with
+less-popular browsers possibly never gaining any visibility.
+
+Hence, for analytics purposes, it is better to treat the brands list
+as a unit, and compare it to known lists of brands sent by various
+the various (browser, version) pairs that are to be distinguished.
+This will necessitate regular updates to the list of known lists of
+brands when new browser versions are released or new browsers become
+popular, or else everything will get bucketted as an unknown browser.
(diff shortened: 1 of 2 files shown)


miketaylr:
Thanks for the PR! When you get a chance, can you take care of https://labs.w3.org/repo-manager/pr/id/WICG/ua-client-hints/197?

yoavweiss:
@quasicomputational - can you join the WICG to appease the IPR bots?

miketaylr:
Looking again at the IPR requirements:

> if the contribution does not concern a normative part of a specification, or is editorial in nature (e.g. fixing typos or examples), the contribution can be marked as non-substantive with the button above - this requires to be logged-in in this system.

I think this falls into the former (non-normative additions), as its adding some additional SHOULD-level advice to the explainer. I'll mark it as such, then can circle back and make the recommend changes to the spec.

miketaylr (approved):
README stuff LGTM, @yoavweiss can you review the tiny patch I added to the spec?

README.md
```diff
@@ -304,6 +304,32 @@ Sites that wish to provide market share analytics using UA-CH will need to
 inspect the `Sec-CH-UA` header, that is sent by default on every request, and
 keep a record of it.
 
+By design, looking at individual entries in the brands list makes it
+hard to distinguish between a less-popular browser's truthful brand
+name and a more-popular browser's arbitrary GREASE. Since the
+less-popular browser may include several popular brand names for
+compatibility purposes, its users will likely be bucketted as using
```
yoavweiss, line 311: s/bucketted/bucketed/ ?

yoavweiss (approved):
LGTM % typo (I think..)

README.md
miketaylr: oh yeah, thanks for catching that.

Issue #236, opened 2021-05-25
Clarify the meaning of 'network' 


miketaylr:
In https://wicg.github.io/ua-client-hints/#fingerprinting:
> The primary goal of User Agent Client Hints is to reduce the default entropy available to the network for passive fingerprinting.

"Default entropy" and "network" might be slightly ambiguous here, so let's try to improve it.

miketaylr:
(maybe ambiguous is the wrong word, unclear?)

My understanding of the terms:

default entropy - the bits of identifying information that could be used to create a fingerprint from the User-Agent string, which is sent by default for all requests

network - here we're referring to the HTTP layer, in particular the `User-Agent` header (see https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3) that is sent over HTTP(S)

Anybody have any suggested improvements to the wording here?

ronancremin:
"Network" usually means a group of connected things. The phrase "available to the network" implies the ability to observe HTTP requests by elements in the network but in a HTTPS world only the destination server sees the contents of the request, so network doesn't feel like the right word when we're talking visibility of information for fingerprinting.  

I think that saying "HTTP layer" actually confuses things further since "layer" suggests the OSI network layer model in which HTTP is actually at the application layer. 

For consistency and alignment, why not use the term used in the HTTP protocol RFCs: "server" or "HTTP server"?

miketaylr:
> "Network" usually means a group of connected things. The phrase "available to the network" implies the ability to observe HTTP requests by elements in the network but in a HTTPS world only the destination server sees the contents of the request, so network doesn't feel like the right word when we're talking visibility of information for fingerprinting.

That's fair. In my mind you have a 1st party site as the root node, and any 3rd party subresources as connected nodes that might also be capture this info. But I don't think it's crucial to capture the connected-ness. 

> For consistency and alignment, why not use the term used in the HTTP protocol RFCs: "server" or "HTTP server"?

"to a server" or "to servers" is a good suggestion. Want to send in a PR?

ronancremin:
Will do.

PR #239, opened 2021-05-31
Language clarification

ronancremin:
Clarify language around entities & passive fingerprinting

***
<a href="https://pr-preview.s3.amazonaws.com/ronancremin/ua-client-hints/pull/239.html" title="Last updated on May 31, 2021, 12:38 PM UTC (6479f38)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/239/017499a...ronancremin:6479f38.html" title="Last updated on May 31, 2021, 12:38 PM UTC (6479f38)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -592,7 +592,7 @@ Fingerprinting {#fingerprinting}
 --------------
 
 The primary goal of User Agent Client Hints is to reduce the default entropy
-available to the network for [=passive fingerprinting=]. However, it will still be possible
+available to a server for [=passive fingerprinting=]. However, it will still be possible
 for some, or all, hints to be requested and used for [=active fingerprinting=] purposes by
 first or delegated third parties. As noted in [[#access]], [=User agents=] should consider
 policies to restrict or reduce access to parties that are known to <a lt="active fingerprinting">


yoavweiss:
Hey @ronancremin!

I'll let @miketaylr weigh in, but IMO, "the network" and "the server" are not interchangeable here. The former also explains why e.g. this feature is limited to secure contexts, while the latter does not.

ronancremin:
We discussed this in #236. The phrase "available to the network" doesn't feel right given that UA-CH will only be sent in secure contexts where network elements other than the destination server should not be able to see request payloads in any case.

amtunlimited:
> UA-CH will only be sent in secure contexts

No real opinion on the wording here, but I think the " default entropy available" referred to by 

> reduce the default entropy available to the network for [=passive fingerprinting=]

Is referring to the current status of the `User-Agent` request header, which is available on all requests

ronancremin:
Agreed, but in a HTTPS world the entropy is not available to _network_, it's only available to the destination server.

miketaylr:
> but IMO, "the network" and "the server" are not interchangeable here. The former also explains why e.g. this feature is limited to secure contexts, while the latter does not.

@yoavweiss can you elaborate? I'm not sure I follow.

> Agreed, but in a HTTPS world the entropy is not available to the _network_, it's only available to the destination server.

There may be more than one destination server though, right? (In the case of a a page with 3rd party subresources... but by default they won't get certain hints unless delegated to).

That said, I'm not sure the difference between network and server is super meaningful, but I'm curious to hear more from @yoavweiss before landing this.

yoavweiss:
Ignore me, I just realized that there's a full section above talking about network based attacks, and why we're limiting the feature to secure contexts above the one that's modified. I'm fine with this change.

miketaylr:
No worries, thanks for taking a look!

miketaylr (approved):
LGTM, thanks @ronancremin

PR #243, opened 2021-06-10
Fix typos

albertodeago:
I think those are a couple of markdown typos (unmatched backticks)


yoavweiss (approved):
Thanks for catching that and fixing!

PR #245, opened 2021-06-29
Fixes #220: More fully specify Sec-CH-UA-Platform-Version

### LINKED ISSUE #220: On Windows, platform version string should be sufficient to differentiate between the various Win10+ releases
Windows historically updated the major or minor version number with each significant OS release. As a result, the "Windows NT <major>.<minor>" component of the UA string was sufficient for sites to know how to send appropriate installers targeting each OS release.

As of Windows 10, though, Windows has essentially frozen the major and minor build number. As a result, this logic is no longer able to work across different Win10 releases even as we approach the 6th anniversary of the first Win10 release.

For OS X, we're already exposing a similar level of detail in terms of version numbers across releases.

With the UA string, it would have been unsafe from a compat perspective to add a third version component since it would have broken many existing UA string parsers. With UA Client Hints, we have an opportunity to add this functionality back.

There's obviously a bit more entropy here, but this version component has only been changing about twice a year and with the Windows install base it's not particularly unique. Since this isn't a header included by default, we can also consider it as an input into any future "privacy budget" schemes.

So, in short: I think we should have browsers expose major.minor.build in `Sec-CH-UA-Platform-Version` on Windows clients. e.g. instead of "10.0" it would say something like "10.0.19042" (for "20H2").

This is possibly not really an issue to resolve here on the spec, but if we have general agreement we could at have at least one of the examples demonstrate this.

erik-anderson:
Document expected version formats by platform based on the currently shipping implementation in Chromium.

On Windows, we diverge from the current shipping implementation by defining it in terms of the `Windows.Foundation.UniversalApiContract` per the discussion in #220.

***
<a href="https://pr-preview.s3.amazonaws.com/erik-anderson/ua-client-hints/pull/245.html" title="Last updated on Jun 30, 2021, 8:01 PM UTC (8eb970f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/245/20da7bc...erik-anderson:8eb970f.html" title="Last updated on Jun 30, 2021, 8:01 PM UTC (8eb970f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -622,7 +622,7 @@ information about the platform version on which a given [=user agent=] is execut
             1. Parse |platformReturnedVersionString|, storing the first three decimal integers delimited by U+002E (FULL STOP). If |platformReturnedVersionString| does not contain three integer values, replace the missing components with "0".
             1. Append all three version components in the order they were read to |platformVersionComponentList|.
         * If the platform is macOS:
-            1. Let |platformReturnedVersionString| be the result of querying the `UIDevice` return by `currentDevice` and reading its `systemVersion`.
+            1. Let |platformReturnedVersionString| be the result of querying the `UIDevice` returned by `currentDevice` and reading its `systemVersion`.
             1. Parse |platformReturnedVersionString|, storing the first three decimal integers delimited by U+002E (FULL STOP). If |platformReturnedVersionString| does not contain three integer values, replace the missing components with "0".
             1. Append all three version components in the order they were read to |platformVersionComponentList|.
         * If the platform is Linux:
--- a/index.bs
+++ b/index.bs
@@ -169,7 +169,7 @@ In response, the user agent includes the platform version information in the nex
   Sec-CH-UA: "Examplary Browser"; v="73", ";Not?A.Brand"; v="27"
   Sec-CH-UA-Mobile: ?0
   Sec-CH-UA-Platform: "Windows"
-  Sec-CH-UA-Full-Version: "14"
+  Sec-CH-UA-Full-Version: "14.0.0"
 ```
 
 ## Use Cases ## {#use-cases}
@@ -630,9 +630,11 @@ information about the platform version on which a given [=user agent=] is execut
             1. Let |platformReturnedVersionString| be the result of querying the `NSProcessInfo` returned by `processInfo` and reading its `operatingSystemVersion`.
             2. Append the `majorVersion`, `minorVersion`, and `patchVersion` components (in that order) to |platformVersionComponentList|.
         * If the platform is Windows:
-            1. If available (i.e., on Windows 10 or higher), append the single integer version of the `Windows.Foundation.UniversalApiContract` contract to |platformVersionComponentList|. Otherwise, append 0 to |platformVersionComponentList|.
-        * Other platforms not listed above:
+            1. If available (i.e., on Windows 10 or higher), let |platformReturnedVersionString| be the result of querying the `Windows.Foundation.UniversalApiContract` integer version and converting it to a string. Otherwise, let |platformReturnedVersionString| be "0".
+            1. Let |platformVersionComponentList| be the result of running <a lt="parse a platform-returned version string">parse a platform-returned version string</a> with |platformReturnedVersionString|.
+        * Otherwise:
             1. Append one to three version parts based on the a format likely to lead to interoperability with other browsers running on that platform to |platformVersionComponentList|.
+            2. While |platformVersionComponentList|'s length is less than 3, append "0" to |platformVersionComponentList|.
     1. Return the result of the [=concatenation=] of |platformVersionComponentList| with a U+002E FULL STOP (`.`) separator.
 
 <p>The <dfn>parse a platform-returned version string</dfn> algorithm, given a string |input|, runs these steps:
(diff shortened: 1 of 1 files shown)


erik-anderson:
@miketaylr I haven't installed the Bikeshed tooling so there may be some formatting things that need to be cleaned up.

miketaylr:
> @miketaylr I haven't installed the Bikeshed tooling so there may be some formatting things that need to be cleaned up.

👍  Formatting looks good, per https://pr-preview.s3.amazonaws.com/erik-anderson/ua-client-hints/pull/245.html

index.bs
```diff
@@ -614,6 +614,30 @@ information about the platform version on which a given [=user agent=] is execut
 [=Structured Header=] whose value MUST be a [=structured header/string=]
 [[!RFC8941]].
 
+[=User agents=] SHOULD return a version with a platform-specific format that allows differentiating significant platform versions by running these steps:
+    1. Let |platformVersionComponentList| be a [=list=].
+    1. Use per-platform logic:
+        * If the platform is Android:
+            1. Let |platformReturnedVersionString| be the result of querying the OS's `android.os.Build.VERSION.RELEASE` string.
+            1. Parse |platformReturnedVersionString|, storing the first three decimal integers delimited by U+002E (FULL STOP). If |platformReturnedVersionString| does not contain three integer values, replace the missing components with "0".
```
miketaylr, line 622: Since this is used in 3 places, it's probably useful to pull out a `"Parse the platform returned version string"` `<dfn>` here (modulo a better name), which can take a `platformReturnedVersionString` as an argument (presumably https://infra.spec.whatwg.org/#strictly-split (on 0x2E), then do some kind of `parseInt` operation (maybe?), grab the first 3 ints or zeros otherwise). 

If it returns a list of the 3 version components we care about (or 0s), you could then set `platformVersionComponentList` to the return value in the next step.

index.bs
```diff
@@ -614,6 +614,30 @@ information about the platform version on which a given [=user agent=] is execut
...
+            1. Append all three version components in the order they were read to |platformVersionComponentList|.
+        * If the platform is macOS:
+            1. Let |platformReturnedVersionString| be the result of querying the `UIDevice` return by `currentDevice` and reading its `systemVersion`.
```
miketaylr: (Is there a missing word here? Query the `UIDevice` string returned by? Or maybe just an expression I'm not familiar with)

index.bs
```diff
@@ -614,6 +614,30 @@ information about the platform version on which a given [=user agent=] is execut
...
+            1. Parse |platformReturnedVersionString|, storing the first three decimal integers delimited by U+002E (FULL STOP). If |platformReturnedVersionString| does not contain three integer values, replace the missing components with "0".
+            1. Append all three version components in the order they were read to |platformVersionComponentList|.
+        * If the platform is Linux:
+            1. Let |platformReturnedVersionString| be the result of querying the `release` string in the `utsname` struct returned by the `uname` API.
+            1. Parse |platformReturnedVersionString|, storing the first three decimal integers delimited by U+002E (FULL STOP). If |platformReturnedVersionString| does not contain three integer values, replace the missing components with "0".
+            1. Append all three version components in the order they were read to |platformVersionComponentList|.
+        * If the platform is macOS:
+            1. Let |platformReturnedVersionString| be the result of querying the `NSOperatingSystemVersion`.
```
miketaylr, line 633: nit: missing `API` at end of sentence?

miketaylr (changes requested):
Thanks for the PR! Looks good, just a few suggested tweaks.

@erik-anderson also, please add your name to https://wicg.github.io/ua-client-hints/#ack (if you'd like).

index.bs
```diff
@@ -614,6 +614,30 @@ information about the platform version on which a given [=user agent=] is execut
...
 [[!RFC8941]].
 
+[=User agents=] SHOULD return a version with a platform-specific format that allows differentiating significant platform versions by running these steps:
+    1. Let |platformVersionComponentList| be a [=list=].
+    1. Use per-platform logic:
+        * If the platform is Android:
+            1. Let |platformReturnedVersionString| be the result of querying the OS's `android.os.Build.VERSION.RELEASE` string.
+            1. Parse |platformReturnedVersionString|, storing the first three decimal integers delimited by U+002E (FULL STOP). If |platformReturnedVersionString| does not contain three integer values, replace the missing components with "0".
+            1. Append all three version components in the order they were read to |platformVersionComponentList|.
+        * If the platform is macOS:
```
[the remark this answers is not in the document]
erik-anderson: Oops-- yep!

index.bs
erik-anderson: I tried to review other Bikeshed docs and didn't see much about parsing things-- I think it's likely fine to just call out that it must be an unsigned integer.

index.bs
erik-anderson: I made the pointer more clear to the API that Chromium is using.

index.bs
erik-anderson: I made the pointer more clear to the API that Chromium is using and read more like the iOS description.

miketaylr:
> I believe this aligns with what's currently in Chrome Canary. Previously the platform version number format was a little all over the place which https://chromium-review.googlesource.com/c/chromium/src/+/2891972 recently changed to always report three version components.

Oh hey, I somehow missed that landing 😅 .

> Would you prefer me to update the Windows description to have three parts (with the 1st component being the UniversalApiContract version and the 2nd and 3rd being 0)?

I think there's a pretty good argument for consistency if we go that route. WDYT?

miketaylr (approved):
LGTM! I'll defer to you if you think we should also require 3 components for Windows (and hold off merging until you let me know). If so, we should update the "one to three version parts" for "Other platforms" to be consistent.

erik-anderson:
@miketaylr I think consistency for parsing across platforms is reasonable. We'll still ensure reasonable version comparison checks on Windows by sticking with only filling in the significant most version component. Please merge if the latest commit looks good.

miketaylr (approved):
LGTM, thanks!

PR #250, opened 2021-08-11
Fixes #196 - Add a Sec-CH-UA-Full-Version-List client hint

### LINKED ISSUE #196: UA full version info is too tightly bound to primary brand instead of engine
Imagine you're a browser embedding an engine that is more broadly used by a larger browser vendor.

In rare occasions, sites may need to work around a platform bug which is later addressed within a major version patch of the shared engine. In this scenario, let's assume it's not otherwise detectable/observable outside of version number.

The way a site would likely get the context for that with UA-Client-Hints is to request the full version via `Accept-CH: Sec-CH-UA-Full-Version` or calling `navigator.userAgentData.getHighEntropyValues(["uaFullVersion"])`, parsing it, and then scoping the workaround to builds without the bug fix.

It's also likely the site will do that targeting solely on the version number of the largest browser shipping that engine. At this point, we've lost a lot of the utility of having a fuller set of brands available via the `sec-ch-ua` header and the navigator.userAgentData.brands list.

To make this more concrete, here's a current example Microsoft Edge UA string:
`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36 Edg/88.0.705.63`

If Edge were to turn on UA Client Hints, Chrome would report `88.0.4324.150` for the UA full version and Edge would report `88.0.705.63`.

Let's say there's a an issue in Chromium versions <= to that version. Even if someone understands it's a Chromium issue affecting multiple browsers and gates their check on "brands _contains_ Chromium", they would need to be aware of every browser's version that maps to that Chromium release. Instead, they might naively author a check like `if (majorVersion == 88 && minor == 0 && build <= 4324) { /* tell user to update their browser */ }` and break non-Chrome embedders that choose a different versioning scheme.

A browser like Edge could choose to simply report the Chromium version number for the full UA version, but that then precludes a site matching on Edge version number (which would be unfortunate since Edge could have its own bugs that need such targeting).

I think an approach that looks something like this would be more likely to help sites author the correct checks:
* The base brand version string should report a number with a version format that matches the full version, but zeroed out for the non-major version numbers. i.e. it might look like `sec-ch-ua: " Not A;Brand";v="99.0.0.0", "Chromium";v="88.0.0.0", "Microsoft Edge";v="88.0.0.0"`.

   If full UA version info is requested via `Accept-CH`, all of the other version parts would be filled in, e.g. `sec-ch-ua: " Not A;Brand";v="99.1.300.grease", "Chromium";v="88.0.4324.150", "Microsoft Edge";v="88.0.705.63"`. Someone parsing the string would now be able to directly target a specific Chromium release across different browser brands.

* For JS callers, the `navigator.userAgentData.brands`' version value would work the same way-- placeholder version numbers for consistent parsing. Perhaps also expose a boolean property (`navigator.uaData.fullVersionAvailable`?) that tells you if the version info is truncated so that version checks don't need to check if all of the non-major version components are zeroed out.

   `getHighEntropyValues` for a `uaFullVersion` request might return a brands value with the full version info now filled out and future calls to `navigator.userAgentData.brands` would also show the full version info. I would prefer to not have a single string passed provided as currently defined due to the aforementioned potential "brand-versus-platform version" confusion it introduces for site owners.

miketaylr:
TODO: 
* [x] deprecate `Sec-CH-UA-Full-Version`
* [x] should getHighEntropyHints return a FrozenArray for valueList?
* [x] update README
* [x] file corresponding CH Infra issue (to capture permissions policy), and coordinate landing that PR once this is ready
* [x] https://wicg.github.io/ua-client-hints/#create-an-arbitrary-version needs to be updated to take an argument as well

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 13, 2021, 9:15 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fmiketaylr%2Fua-client-hints%2F8cc707f3fb3a11ed9ed645781c7bfeb1fd989a17%2Findex.bs&md-warning=not%20ready)

```
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ua-client-hints%23250.)._
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ accomplish this as follows:
       readonly attribute FrozenArray<NavigatorUABrandVersion> brands; // [ {brand: "Google Chrome", version: "84"}, {brand: "Chromium", version: "84"} ]
       readonly attribute boolean mobile; // false
       readonly attribute platform; // "PhoneOS"
-      Promise<UADataValues> getHighEntropyValues(sequence<DOMString> hints); // [ "arm", "64", "X644GTM", "PhoneOS", "10A", "73.32.AGX.5" ]
+      Promise<UADataValues> getHighEntropyValues(sequence<DOMString> hints); // { architecture: "arm", bitness: "64", model: "X644GTM", platform: "PhoneOS", platformVersion: "10A", versionList: [ {brand: "Google Chrome", version: "84.1.2.3"}, {brand: "Chromium", version: "84.1.2.3"}, {brand: "Not A;Brand", version: "101.3.2.9"} ] }
     };
 
     interface mixin NavigatorUA {
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ accomplish this as follows:
         brand list. For example:
 
         ```http
-        Sec-CH-UA-Version-List: "Edge"; v="92.0.902.73", "Chromium"; v="92.0.4515.131", "?Not:Your Browser"; v="3.1.2.0"
+        Sec-CH-UA-Version-List: "Microsoft Edge"; v="92.0.902.73", "Chromium"; v="92.0.4515.131", "?Not:Your Browser"; v="3.1.2.0"
         ```
 
 4.  These client hints should also be exposed via JavaScript APIs via a new
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ a corresponding JavaScript API:
 * `Sec-CH-UA-Platform`
 * `Sec-CH-UA-Platform-Version`
 * `Sec-CH-UA`
-* `Sec-CH-UA-Full-Version` (deprecated in favor of `Sec-CH-UA-Version-List`)
-* `Sec-CH-UA-Version-List`
+* `Sec-CH-UA-Full-Version` (deprecated in favor of `Sec-CH-UA-Full-Version-List`)
+* `Sec-CH-UA-Full-Version-List`
 
 
 ## Contributing
@@ -221,7 +221,7 @@ accomplish this as follows:
         ```
 
         Advisement: `Sec-CH-UA-Full-Version` is deprecated and will be removed in the future.
-        Developers should use `Sec-CH-UA-Version-List` instead.
+        Developers should use `Sec-CH-UA-Full-Version-List` instead.
 
     1.  The `Sec-CH-UA-Platform` header field represents the platform's brand and major version. For
         example:
@@ -230,11 +230,11 @@ accomplish this as follows:
         Sec-CH-UA-Platform: "Windows"
         ```
 
-    1.  The `Sec-CH-UA-Version-List` header field represents the full version for each brand in its
+    1.  The `Sec-CH-UA-Full-Version-List` header field represents the full version for each brand in its
         brand list. For example:
 
         ```http
-        Sec-CH-UA-Version-List: "Microsoft Edge"; v="92.0.902.73", "Chromium"; v="92.0.4515.131", "?Not:Your Browser"; v="3.1.2.0"
+        Sec-CH-UA-Full-Version-List: "Microsoft Edge"; v="92.0.902.73", "Chromium"; v="92.0.4515.131", "?Not:Your Browser"; v="3.1.2.0"
         ```
 
 4.  These client hints should also be exposed via JavaScript APIs via a new
@@ -252,11 +252,11 @@ accomplish this as follows:
       boolean mobile;             // true
       DOMString architecture;     // "arm"
       DOMString bitness;          // "64"
+      FrozenArray&lt;NavigatorUABrandVersion&gt; fullVersionList; // [ {brand: "Google Chrome", version: "84.0.4147.0"}, {brand: "Chromium", version: "84.0.4147"} ]
       DOMString model;            // "X644GTM"
       DOMString platform;         // "PhoneOS"
       DOMString platformVersion;  // "10A"
-      DOMString uaFullVersion; // deprecated in favor of versionList
-      FrozenArray&lt;NavigatorUABrandVersion&gt; versionList; // [ {brand: "Google Chrome", version: "84.0.4147.0"}, {brand: "Chromium", version: "84.0.4147"} ]
+      DOMString uaFullVersion; // deprecated in favor of fullVersionList
     };
 
     [Exposed=(Window,Worker)]
@@ -264,7 +264,7 @@ accomplish this as follows:
       readonly attribute FrozenArray<NavigatorUABrandVersion> brands; // [ {brand: "Google Chrome", version: "84"}, {brand: "Chromium", version: "84"} ]
       readonly attribute boolean mobile; // false
       readonly attribute platform; // "PhoneOS"
-      Promise<UADataValues> getHighEntropyValues(sequence<DOMString> hints); // { architecture: "arm", bitness: "64", model: "X644GTM", platform: "PhoneOS", platformVersion: "10A", versionList: [ {brand: "Google Chrome", version: "84.1.2.3"}, {brand: "Chromium", version: "84.1.2.3"}, {brand: "Not A;Brand", version: "101.3.2.9"} ] }
+      Promise<UADataValues> getHighEntropyValues(sequence<DOMString> hints); // { architecture: "arm", bitness: "64", model: "X644GTM", platform: "PhoneOS", platformVersion: "10A", fullVersionList: [ {brand: "Google Chrome", version: "84.1.2.3"}, {brand: "Chromium", version: "84.1.2.3"}, {brand: "Not A;Brand", version: "101.3.2.9"} ] }
     };
 
     interface mixin NavigatorUA {
@@ -306,7 +306,7 @@ Sec-CH-UA-Platform: “Windows”
 If a server delivers the following response header:
 
 ```http
-Accept-CH: Sec-CH-UA-Version-List, Sec-CH-UA-Arch
+Accept-CH: Sec-CH-UA-Full-Version-List, Sec-CH-UA-Arch
 ```
 
 Then subsequent requests to `https://example.com` will include the following request headers:
@@ -317,7 +317,7 @@ User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML,
 Sec-CH-UA: "Chrome"; v="74", ";Not)Your=Browser"; v="13"
 Sec-CH-UA-Mobile: ?0
 Sec-CH-UA-Platform: "Windows"
-Sec-CH-UA-Version-List: "Chrome"; v="74.0.3729.0", "Chromium"; v="74.0.3729.0", "?Not:Your Browser"; v="13.0.1.0"
+Sec-CH-UA-Full-Version-List: "Chrome"; v="74.0.3729.0", "Chromium"; v="74.0.3729.0", "?Not:Your Browser"; v="13.0.1.0"
 Sec-CH-UA-Arch: "arm"
 ```
 
@@ -335,12 +335,12 @@ Sec-CH-UA-Mobile: ?0
 Sec-CH-UA-Platform: “Windows”
 ```
 
-The server responds that the `Sec-CH-UA-Version-List` is required on first-request in order to
+The server responds that the `Sec-CH-UA-Full-Version-List` is required on first-request in order to
 deliver some optimized resource, for example:
 
 ```http
-Accept-CH: Sec-CH-UA-Version-List
-Critical-CH: Sec-CH-UA-Version-List
+Accept-CH: Sec-CH-UA-Full-Version-List
+Critical-CH: Sec-CH-UA-Full-Version-List
 ```
 
 The client then retries the initial request with the requested hints:
@@ -351,7 +351,7 @@ User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML,
 Sec-CH-UA: "Chrome"; v="74", ";Not)Your=Browser"; v="13"
 Sec-CH-UA-Mobile: ?0
 Sec-CH-UA-Platform: “Windows”
-Sec-CH-UA-Version-List: "Chrome"; v="74.0.3729.0", "Chromium"; v="74.0.3729.0", "?Not:Your Browser"; v="13.0.1.0"
+Sec-CH-UA-Full-Version-List: "Chrome"; v="74.0.3729.0", "Chromium"; v="74.0.3729.0", "?Not:Your Browser"; v="13.0.1.0"
 ```
 
 The user agent can make reasonable decisions about when to honor requests for detailed user agent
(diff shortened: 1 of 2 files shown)


miketaylr:
@erik-anderson mind providing some feedback on this PR?

yoavweiss:
High-level question on this: Initially, @mikewest defined `Sec-CH-UA` so that when an opt-in is provided, its processing fills in the full version instead of the significant version. We later removed that in favor of `Full-Version`, to make processing simpler (to not special case `Sec-CH-UA`). Maybe that was a mistake?

mikewest:
> High-level question on this: Initially, @mikewest defined `Sec-CH-UA` so that when an opt-in is provided, its processing fills in the full version instead of the significant version. We later removed that in favor of `Full-Version`, to make processing simpler (to not special case `Sec-CH-UA`). Maybe that was a mistake?

I still like the original version. :)

But given the way things have evolved, I'm not sure it's worth rolling everything back. Chrome has shipped `Sec-CH-UA` for a while now, and changing its semantics might be difficult.

miketaylr:
> Initially, @mikewest defined `Sec-CH-UA` so that when an opt-in is provided, its processing fills in the full version instead of the significant version. We later removed that in favor of `Full-Version`, to make processing simpler (to not special case `Sec-CH-UA`).

Hmm, that's pretty interesting, I think I missed out on that entirely! (Or knew at one point and just forgot... 🙈 ).

Recapping for myself: the early design was if you requested `ua` (now it would have to be `sec-ch-ua`) you would get a full-versioned brandlist, rather than the default significant versioned one.

I think folks might request `sec-ch-ua` to get at (full!) versions without understanding that they get significant versions for free, so a different-enough token would probably be desirable were we to take that approach. Maybe `sec-ch-ua-full-version(s)` could have powered that behavior... but that ship has sailed. I don't think we want to make stumbling into requesting extra entropy to be so easy. It wouldn't' surprise me to find code where folks request the full version list and parse it to get at a significant version.

And looking back to https://github.com/WICG/ua-client-hints/issues/45, I can see @yoavweiss had the same concerns. WDYT, @yoavweiss? Do we go back to the magic behavior with a different token (we could re-use this proposals`sec-ch-ua-version-list`), or do we introduce a new hint?

One one hand, there's fewer headers over the wire if you want full-version via `sec-ch-ua`, on the other, it's kind of magical or confusing.

Also interested in feedback from @erik-anderson.

yoavweiss:
> One one hand, there's fewer headers over the wire if you want full-version via `sec-ch-ua`, on the other, it's kind of magical or confusing.

Yeah, I agree with that description of the tension. Given that requesting for the full version should be rather rare, let's go with less magic and confusion.

README.md
```diff
@@ -12,7 +12,8 @@ a corresponding JavaScript API:
 * `Sec-CH-UA-Platform`
 * `Sec-CH-UA-Platform-Version`
 * `Sec-CH-UA`
-* `Sec-CH-UA-Full-Version`
+* `Sec-CH-UA-Full-Version` (deprecated in favor of `Sec-CH-UA-Version-List`)
```
yoavweiss, line 15: Do we have use counters for `Full-Version`? Could we just change its semantics in place (instead of rename)?

miketaylr: Let me check on that. It would be nicer if we could...

miketaylr: OK, we might have missed the boat here with 1.6% of page loads, sadly 😿:

https://chromestatus.com/metrics/feature/timeline/popularity/3189

Screen Shot 2021-08-18 at 1 46 53 PM

miketaylr: @yoavweiss looks like it's too high to change here. Shall we continue review assuming a new hint?

mikewest: 1.6% is an astoundingly high number. Do you have any insight into who's using this mechanism? Is there some large embedded resource that's causing this to show up on lots of sites, or are there large top-level sites opting in?

index.bs
```diff
@@ -485,23 +485,19 @@ The header's ABNF is:
...
-
-1.  Let |list| be a [=/list=], initially empty.
-
-2. For each |brandVersion| in [=user agent/brands=]:
+To <dfn abstract-op>return the `Sec-CH-UA` value for a request</dfn>,
+[=user agents=] MUST:
 
+1. Let |list| be a [=/list=], initially empty.
+1. Let |brands| be the result of running the [=create brands=] steps, with
+    [=user agent/significant version=] passed as an argument.
```
yoavweiss, line 493: It's unclear if what is passed is the version itself, or an instruction to operate on that version type. I think you meant the latter, but it might be worthwhile to pass an internal enum, and then refer internally to the right version type based on that enum.

Also, right now both significant version and full version are defined as a single version. It may be worthwhile to adapt their descriptions to indicate that, unlike Highlander, there can be more than one.

index.bs
```diff
@@ -485,23 +485,19 @@ The header's ABNF is:
...
-    2. Set |parameter|["param_name"] to "v".
-
-    3. Set |parameter|["param_value"] to |brandVersion|'s {{NavigatorUABrandVersion/version}}.
-
-    2. Let |pair| be a tuple comprised of |brandVersion|'s {{NavigatorUABrandVersion/brand}} and |parameter|.
-
-    3. Append |pair| to |list|.
-
-3. Return the output of running [=serializing a list=] with |list| as input.
+    1. Set |parameter|["param_name"] to "v".
```
yoavweiss, line 496: I think the serialization can be moved to a helper algorithm as well, to avoid duplicating it in both places.

index.bs
```diff
@@ -671,14 +697,15 @@ dictionary NavigatorUABrandVersion {
 };
 
 dictionary UADataValues {
-  sequence&lt;NavigatorUABrandVersion&gt; brands;
+  FrozenArray&lt;NavigatorUABrandVersion&gt; brands;
```
yoavweiss, line 700: Does that mean that we need to test identity in implementations of the references returned by multiple calls to getHighEntropyValues? Will the Chromium implementation need to align here?

README.md
miketaylr: We know that YouTube has been experimenting with UA-CH (and sadly [we caused a crash there, oops](https://bugs.chromium.org/p/chromium/issues/detail?id=1217589#c17)), and we have some other partners who have committed to doing some testing.

README.md
```diff
@@ -237,20 +248,23 @@ accomplish this as follows:
...
+      FrozenArray&lt;NavigatorUABrandVersion&gt; versionList; // [ {brand: "Google Chrome", version: "84.0.4147.0"}, {brand: "Chromium", version: "84.0.4147"} ]
     };
 
     [Exposed=(Window,Worker)]
     interface NavigatorUAData {
       readonly attribute FrozenArray<NavigatorUABrandVersion> brands; // [ {brand: "Google Chrome", version: "84"}, {brand: "Chromium", version: "84"} ]
       readonly attribute boolean mobile; // false
       readonly attribute platform; // "PhoneOS"
-      Promise<UADataValues> getHighEntropyValues(sequence<DOMString> hints); // { "arm", "64", "X644GTM", "PhoneOS", "10A", "73.32.AGX.5" }
+      Promise<UADataValues> getHighEntropyValues(sequence<DOMString> hints); // [ "arm", "64", "X644GTM", "PhoneOS", "10A", "73.32.AGX.5" ]
```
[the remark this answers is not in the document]
miketaylr: Thanks - I think `{ architecture: "arm", bitness: "64", ... }.` is what we're going for... I don't even know what I was thinking when I changed this. Will update.

README.md
```diff
@@ -219,13 +220,23 @@ accomplish this as follows:
...
 
         ```http
         Sec-CH-UA-Platform: "Windows"
         ```
 
+    1.  The `Sec-CH-UA-Version-List` header field represents the full version for each brand in its
+        brand list. For example:
+
+        ```http
+        Sec-CH-UA-Version-List: "Edge"; v="92.0.902.73", "Chromium"; v="92.0.4515.131", "?Not:Your Browser"; v="3.1.2.0"
```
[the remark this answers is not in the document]
miketaylr: Thanks, will fix.

miketaylr:
> This addresses the concern I had in issue #196. Thanks!

Thanks for the review! @erik-anderson any thoughts on `Sec-CH-UA-Version-List` vs `Sec-CH-UA-Full-Version-List`? Former is less chars to type, latter is more accurate/correct.

erik-anderson:
> > This addresses the concern I had in issue #196. Thanks!
> 
> Thanks for the review! @erik-anderson any thoughts on `Sec-CH-UA-Version-List` vs `Sec-CH-UA-Full-Version-List`? Former is less chars to type, latter is more accurate/correct.

I don't think omitting it is less accurate, but it is indeed less verbose and may take someone staring at it slightly harder to understand that it provides full version numbers. I think it would be fine to omit the word "Full" and just spec it to return full versions; that's my slight preference.

index.bs
miketaylr: Addressed in https://github.com/WICG/ua-client-hints/pull/250/commits/01cc7d30ef0ab38952cfdea275dcd7504a8f3aa4.

index.bs
miketaylr: > it might be worthwhile to pass an internal enum, and then refer internally to the right version type based on that enum

This sounds probably like what we want to do, but I'm not sure how to spec it. Are you aware of any other similar usage in a spec I can learn from, or have some suggestions?

README.md
```diff
@@ -12,7 +12,8 @@ a corresponding JavaScript API:
 * `Sec-CH-UA-Platform`
 * `Sec-CH-UA-Platform-Version`
 * `Sec-CH-UA`
-* `Sec-CH-UA-Full-Version`
+* `Sec-CH-UA-Full-Version` (deprecated in favor of `Sec-CH-UA-Version-List`)
+* `Sec-CH-UA-Version-List`
```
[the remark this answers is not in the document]
miketaylr: I think I've decided `Full-Version-List` is the way to go here, even if I don't love how long it is.

index.bs
miketaylr: @yoavweiss does https://github.com/WICG/ua-client-hints/pull/250/commits/ac953438348d12ea2d7ff07d7a0247badfcc0a60 look better from your perspective?

index.bs
```diff
@@ -712,44 +731,47 @@ Processing model {#processing}
...
+<code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is
+initially the result of [=create a frozen array|creating a frozen array=] from the [=user agent=]'s
+[=brands=].
 
-When asked to run the <dfn>create brands</dfn> algorithm, the [=user agent=] MUST run the following steps:
-1. Let |list| be a [=/list=].
+Additionally, every {{WindowOrWorkerGlobalScope}} object has an associated
+<dfn for="WindowOrWorkerGlobalScope">version list frozen array</dfn>, which is a
+<code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is
+initially the result of [=create a frozen array|creating a frozen array=] from the [=user agent=]'s
```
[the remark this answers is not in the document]
miketaylr: Good catch!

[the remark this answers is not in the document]
miketaylr: Fixed in 8ba20836261f6af0adf5a082a3af10f137cee269.

index.bs
```diff
@@ -712,44 +731,47 @@ Processing model {#processing}
...
+Every {{WindowOrWorkerGlobalScope}} object has an associated
+<dfn for="WindowOrWorkerGlobalScope">brands frozen array</dfn>, which is a
+<code><a interface>FrozenArray</a>&lt;<a dictionary>NavigatorUABrandVersion</a>></code>. It is
+initially the result of [=create a frozen array|creating a frozen array=] from the [=user agent=]'s
+[=brands=].
 
-When asked to run the <dfn>create brands</dfn> algorithm, the [=user agent=] MUST run the following steps:
-1. Let |list| be a [=/list=].
+Additionally, every {{WindowOrWorkerGlobalScope}} object has an associated
+<dfn for="WindowOrWorkerGlobalScope">version list frozen array</dfn>, which is a
```
yoavweiss, line 744: Nit: "full version list frozen array" to make it more obvious?

index.bs
```diff
@@ -712,44 +731,47 @@ Processing model {#processing}
...
-
-    2.  Randomize the order of the [=list/items=] in |list|.
+    1. Append |dict| to |list|.
+1.  The [=user agent=] SHOULD execute the following steps:
+    1. [=list/Append=] one additional [=list/item=] to |list| containing a
+        {{NavigatorUABrandVersion}} dictionary, initialized with
+        <a lt="create an arbitrary brand">arbitrary brand</a> as {{NavigatorUABrandVersion/brand}}
+        and the result of <a lt="create an arbitrary version">creating an arbitrary version</a> with
+        {{NavigatorUABrandVersion/version}}.
+    1.  Randomize the order of the [=list/items=] in |list|.
```
yoavweiss, line 767: Should the order in the two lists be the same or could they be randomized to be in different orders? If it should be the same, randomization should be done elsewhere and passed along here. If they can be different, this is fine.

index.bs
```diff
@@ -656,10 +652,32 @@ The header's ABNF is:
...
+The header's ABNF is:
+
+``` abnf
+  Sec-CH-UA-Full-Version-List = sf-list
+```
+
+To <dfn abstract-op>return the `Sec-CH-UA-Full-Version-List` value for a request</dfn>, perform the
+following steps:
+
+1. Let |version type| be the type [=user agent/full version=].
```
yoavweiss, line 671: You're using [=user agent/full version=] here as both the type passed and its value. It's a bit weird.
Maybe instead you can define an internal enum (or string or whatever) that you pass to the algorithm, and then inside the algorithm, split the behavior based on the enum's value?

yoavweiss (commented):
Looks good! A few comments that can make it even gooder

index.bs
miketaylr: I prefer random, to keep things intentional (and not encourage folks to assume orders are stable anywhere).

index.bs
miketaylr: thanks - missed that one :)

index.bs
miketaylr: Thanks - I think I finally get what you mean. Let's land this PR here and I filed https://github.com/WICG/ua-client-hints/issues/254 to fix this up (should be able to do that this week).

soffieswan041:
> TODO:
> 
> * [x]  deprecate `Sec-CH-UA-Full-Version`
> * [x]  should getHighEntropyHints return a FrozenArray for valueList?
> * [x]  update README
> * [x]  file corresponding CH Infra issue (to capture permissions policy), and coordinate landing that PR once this is ready
> * [x]  https://wicg.github.io/ua-client-hints/#create-an-arbitrary-version needs to be updated to take an argument as well
> 
> ###  Error: 400 Bad Request 
> [PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 13, 2021, 9:15 PM UTC)_.
> 
> More

PR #253, opened 2021-08-24
Fixes #248 - Linux returns empty string for Sec-CH-UA-Platform-Version

### LINKED ISSUE #248: Sec-CH-UA-Platform-Version should not return Linux kernel version
Looking at https://wicg.github.io/ua-client-hints/#sec-ch-ua-platform-version I notice:

> If the platform is Linux:
>
> Let platformReturnedVersionString be the result of querying the release string in the utsname struct returned by the uname API.

Suspecting that won't work well, I wrote this test program:

```
#include <errno.h>
#include <stdio.h>
#include <string.h>
#include <sys/utsname.h>

int
main (void)
{
  struct utsname utsname;

  if (uname (&utsname) != 0)
    {
      fprintf (stderr, "utsname failed: %s", strerror (errno));
      return -1;
    }

  printf ("utsname release: %s\n", utsname.release);
  return 0;
}
```

It prints: `utsname release: 5.13.4-200.fc34.x86_64`, which is the Linux kernel version, including all the release components that identify the distribution build revision 200, OS version (Fedora 34), and CPU architecture. For Linux, I would rather leave this hint empty. There's nothing useful that could be returned. Certainly we don't want to expose kernel version ("5.13.4") or OS version ("34") to the web. And you have another hint for CPU architecture.

(For fun, I also checked `utsname.version` and got `#1 SMP Tue Jul 20 20:27:29 UTC 2021`. Not sure who decided that putting the kernel build timestamp in there was a good idea. *Shrug.*)

kyraseevers:
After investigation, the Linux uname API version string does not provide any developer-useful information. It is better to simply return the empty string.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -615,6 +615,8 @@ information about the platform version on which a given [=user agent=] is execut
 [[!RFC8941]].
 
 [=User agents=] SHOULD return a version with a platform-specific format that allows differentiating significant platform versions by running these steps:
+    1. If the platform is Linux:
+    	1. Return the empty string.
     1. Let |platformVersionComponentList| be a [=list=].
     1. Use per-platform logic:
         * If the platform is Android:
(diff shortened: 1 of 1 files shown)


miketaylr:
Thanks for the PR @kyraseevers! Will review today.

> 🚨 [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

This seems unrelated.

index.bs
```diff
@@ -615,6 +615,8 @@ information about the platform version on which a given [=user agent=] is execut
 [[!RFC8941]].
 
 [=User agents=] SHOULD return a version with a platform-specific format that allows differentiating significant platform versions by running these steps:
+    1. If the platform is Linux:
+    	1. Return the empty string.
     1. Let |platformVersionComponentList| be a [=list=].
```
miketaylr, line 620: Unrelated to your commit, but one small suggestion here which might help with readability would be to start this line with "Else, " and perhaps pull "Use per-platform logic" into the sentence. So something like:

`Else, let |platformVersionComponentList| be a [=list=] and use per-platform logic:` 

What do you think @kyraseevers?

miketaylr (commented):
LGTM, but let's wait on IPR bot being happy before merging.

index.bs
kyraseevers: I agree - I think this will help readability significantly. I will implement and re-commit.

yoavweiss:
Hey @kyraseevers! Can you join the WICG to appease the IPR bots?

kyraseevers:
Thanks @yoavweiss! I have put in a request to join WICG, and hopefully it will populate soon.

PR #259, opened 2021-09-20
Revert "Fixes #196 - Add a Sec-CH-UA-Full-Version-List client hint"

### LINKED ISSUE #196: UA full version info is too tightly bound to primary brand instead of engine

soffieswan041:
Reverts WICG/ua-client-hints#250

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 20, 2021, 3:21 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FWICG%2Fua-client-hints%2Fcfb3c39b798fa3cd5b2742a7db8303ba57e8ecd2%2Findex.bs&md-warning=not%20ready)

```
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ua-client-hints%23259.)._
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,7 @@ a corresponding JavaScript API:
 * `Sec-CH-UA-Platform`
 * `Sec-CH-UA-Platform-Version`
 * `Sec-CH-UA`
-* `Sec-CH-UA-Full-Version` (deprecated in favor of `Sec-CH-UA-Full-Version-List`)
-* `Sec-CH-UA-Full-Version-List`
+* `Sec-CH-UA-Full-Version`
 
 
 ## Contributing
@@ -220,21 +219,11 @@ accomplish this as follows:
         Sec-CH-UA-Full-Version: "73.1.2343B.TR"
         ```
 
-        Advisement: `Sec-CH-UA-Full-Version` is deprecated and will be removed in the future.
-        Developers should use `Sec-CH-UA-Full-Version-List` instead.
-
     1.  The `Sec-CH-UA-Platform` header field represents the platform's brand and major version. For
         example:
 
         ```http
         Sec-CH-UA-Platform: "Windows"
-        ```
-
-    1.  The `Sec-CH-UA-Full-Version-List` header field represents the full version for each brand in its
-        brand list. For example:
-
-        ```http
-        Sec-CH-UA-Full-Version-List: "Microsoft Edge"; v="92.0.902.73", "Chromium"; v="92.0.4515.131", "?Not:Your Browser"; v="3.1.2.0"
         ```
 
 4.  These client hints should also be exposed via JavaScript APIs via a new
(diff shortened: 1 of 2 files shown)


miketaylr:
@soffieswan041 hi there, is this intentional? if so, can you open an issue first to discuss? thanks.

PR #265, opened 2021-09-23
Fixes #255 - Make it clear UA-CH doesn't want _more_ entropy out of browsers

### LINKED ISSUE #255: Not clear that this is not increasing (passive) information exposure
After talking to @yoavweiss and @miketaylr about this I learned that the intent here is not to increase the amount of information that browsers expose to sites.  The breadth of information listed in the specification makes it seem like this would be an increase in information for some browsers (Firefox stopped sharing full browser version information a long time ago).  Having information like that listed makes it seem like the intent is to make that information consistently available.

I understand that the intent is to provide the same information as previously, with some information maybe being removed (WOW indicators on Windows, for instance, to start with).  The document should state that up front.

The specification should also make it clear, for each field, that the information will not be provided if browsers don't want to provide it, which might be the case if the information was not previously available.  This is there in [S4.1.6](https://wicg.github.io/ua-client-hints/#getHighEntropyValues), step 2, but it isn't especially prominent.

There are reasons why variations between browser policies for different fields might cause problems, but as this is limited to UA-specific details, it might be acceptable in this case.

Note: Some of our objections to this API are on the basis that the information being provided is both harmful from a privacy perspective and not useful to include.  But we do have to acknowledge that the privacy harms in question already exist and that this is an attempt to find a way to remove the offending information in a managed fashion.  The goal is to decouple a bunch of interconnected compatibility hazards to make that removal easier.  I personally don't think that this gamble will work: people who have a developed a dependency on certain information aren't necessarily inclined to stop using that information, so removal of information is not made easier by this change.  My guess is that the result will only be that more bytes are spent (minimally the three default fields) with the only benefit being a modest improvement clarity and reliability for those who move to using the new syntax.

miketaylr:
PTAL @yoavweiss @martinthomson

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 24, 2021, 2:16 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FWICG%2Fua-client-hints%2Fe60614b4c09f29c74619f97396432c69f123b354%2Findex.bs&md-warning=not%20ready)

```
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ua-client-hints%23265.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -916,12 +916,20 @@ That delegation is defined as part of [=append client hints to request=].
 Fingerprinting {#fingerprinting}
 --------------
 
-The primary goal of User Agent Client Hints is to reduce the default entropy
-available to a server for [=passive fingerprinting=]. However, it will still be possible
-for some, or all, hints to be requested and used for [=active fingerprinting=] purposes by
-first or delegated third parties. As noted in [[#access]], [=User agents=] should consider
-policies to restrict or reduce access to parties that are known to <a lt="active fingerprinting">
-actively fingerprint</a> their users.
+The *primary goal* of User Agent Client Hints is to reduce the amount of default entropy exposed to
+the web at large through the User-Agent header field, which may be used for
+[=passive fingerprinting=] purposes.
+
+[=User agents=] naturally vary in the types and values of information exposed by the UA
+string and some have made deliberate privacy-focused choices to not expose certain bits, such
+as device [=model=] on mobile. This specification makes no requirements for a [=user agent=] to
+expose more entropy via User Agent Client Hints than it does through the User-Agent header field
+today—the empty string is a valid value for all such high-entropy hints.
+
+However, it will still be possible for some, or all, hints to be requested and used for
+[=active fingerprinting=] purposes by first or delegated third parties. As noted in [[#access]],
+[=user agents=] should consider policies to restrict or reduce access to parties that are known to
+<a lt="active fingerprinting">actively fingerprint</a> their users.
 
 Access Restrictions {#access}
 -------------------
--- a/index.bs
+++ b/index.bs
@@ -920,11 +920,10 @@ The *primary goal* of User Agent Client Hints is to reduce the amount of default
 the web at large through the User-Agent header field, which may be used for
 [=passive fingerprinting=] purposes.
 
-[=User agents=] naturally vary in the types and values of information exposed by the UA
-string and some have made deliberate privacy-focused choices to not expose certain bits, such
-as device [=model=] on mobile. This specification makes no requirements for a [=user agent=] to
-expose more entropy via User Agent Client Hints than it does through the User-Agent header field
-today—the empty string is a valid value for all such high-entropy hints.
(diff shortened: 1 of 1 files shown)


yoavweiss (approved):
LGTM

index.bs
```diff
@@ -904,21 +906,30 @@ of a given agent's behavior over time.
...
-actively fingerprint</a> their users.
+The *primary goal* of User Agent Client Hints is to reduce the amount of default entropy exposed to
+the web at large through the User-Agent header field, which may be used for
+[=passive fingerprinting=] purposes.
+
+[=User agents=] naturally vary in the types and values of information exposed by the UA
+string and some have made deliberate privacy-focused choices to not expose certain bits, such
+as device [=model=] on mobile. This specification makes no requirements for a [=user agent=] to
+expose more entropy via User Agent Client Hints than it does through the User-Agent header field
+today—the empty string is a valid value for all such high-entropy hints.
```
martinthomson, line 927: Is an empty string really what you are requiring when a user (or user agent) decides not to share something?

I might phrase this last sentence as:

> A [=user agent=] is able to decide which hints it provides, with the expectation that no more information is provided than would be included in an unredacted User-Agent header field.  A user agent can provide an empty string for any value that it does not wish to provide.

That piece about the empty string needs to be better captured in the specification.  I don't see this anywhere else.  The opposite in fact: the [getHighEntropyValues](https://wicg.github.io/ua-client-hints/#getHighEntropyValues) definition would seem to reject if someone asked for something that the browser (or user) didn't want to share.

index.bs
```diff
@@ -904,21 +906,30 @@ of a given agent's behavior over time.
...
+
+However, it will still be possible for some, or all, hints to be requested and used for
+[=active fingerprinting=] purposes by first or delegated third parties. As noted in [[#access]],
+[=user agents=] should consider policies to restrict or reduce access to parties that are known to
+<a lt="active fingerprinting">actively fingerprint</a> their users.
```
martinthomson, line 932: I'm also seeing a need to strengthen the access restrictions section cited here.  It doesn't concretely say what might happen if a browser decides to restrict access.

index.bs
miketaylr: Thanks for the suggested wording, I like it.

> Is an empty string really what you are requiring when a user (or user agent) decides not to share something?

We have a couple of options: return the hint with a lie, return the hint with the empty string, or refuse to return the header at all (in the case of HTTP anyways, which is I guess functionally equivalent to not supporting the hint). Those all sound like valid choices to me, so I'll tweak the language a bit.

index.bs
miketaylr: Thanks, agree it can be improved. I've filed https://github.com/WICG/ua-client-hints/issues/266 to tackle this.

PR #285, opened 2022-01-18
Fixes #217 - Add a WoW64 boolean hint

### LINKED ISSUE #217: Split up UA-Bitness hint into 2
Thinking more about the use-cases for "bitness" (see #105), it seems desirable to not only know platform bitness, but also the app bitness. For example, if you run an app store webpage, or a support page for an app, you can know if a user has downloaded the most optimized binary, or is in some other "32bit app on 64bit OS" bucket (aka "WOW64"). 

These 2 hints (UA-App-Bitness? / UA-Client-Bitness? & UA-Platform-Bitness) would be high-entropy hints and not available by default.

miketaylr:
PTAL @yoavweiss - this is the last missing piece from the UA string in some browsers (Chromium, perhaps others). The idea is allow for a one to one mapping possible from UA to UA-CH.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -458,6 +458,7 @@ has defined a number of properties for itself:
       "Pixel 2 XL")
 *   <dfn for="user agent" export>mobileness</dfn> - A boolean indicating if the [=user agent=]'s
       device is a mobile device. (e.g., ?0 or ?1)
+*   <dfn for="user agent" export>wow64-ness</dfn> - A boolean indicating if the [=user agent=]'s binary is running in 32-bit mode on 64-bit Windows. (e.g., ?0 or ?1)
 
 
 [=User agents=] SHOULD keep these strings short and to the point, but servers MUST accept arbitrary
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -681,9 +683,22 @@ The header's ABNF is:
   Sec-CH-UA-Platform-Version = sf-string
 ```
 
+The 'Sec-CH-UA-WoW64' Header Field {#sec-ch-ua-wow64}
```
abeyad, line 686: Do we want to have `Sec-CH-App-Bitness` instead that offers a generic hint, instead of windows specific?

Pros:
- It's not something windows-specific
- It could be used for other use cases (e.g. to disable some features on 32 bit binaries on 64 bit platforms)

Cons:
- If we'd rather not offer this as a generic hint, keeping it windows specific to just make the UA->UA-CH transition easier makes sense.  There are other ways for sites to know if they should enable/disable certain features that shouldn't just be tied to 32-bit app on 64-bit platform.

miketaylr: > Do we want to have Sec-CH-App-Bitness instead that offers a generic hint, instead of windows specific?

I think if we had a handful of compelling use cases, I would prefer this. But since we don't, I opted for the safer "pave the backwards compatible cowpath" approach.

abeyad: Makes sense

yoavweiss (approved):
LGTM

miketaylr:
thanks all

PR #288, opened 2022-01-31
Fixes #287 - Return a non-zero value for Win 7, Win 8, Win 8.1

### LINKED ISSUE #287: Platform Version hint doesn't return anything useful for Windows < 10
The market share is dwindling over time, but I think it's cleaner for the API to return something useful (other than "0").

miketaylr:
***
<a href="https://pr-preview.s3.amazonaws.com/miketaylr/ua-client-hints/pull/288.html" title="Last updated on Feb 9, 2022, 12:22 AM UTC (22042e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/288/10bb1df...miketaylr:22042e9.html" title="Last updated on Feb 9, 2022, 12:22 AM UTC (22042e9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -582,8 +582,8 @@ The 'Sec-CH-UA-Model' Header Field {#sec-ch-ua-model}
 -------------------------------
 
 The <dfn http-header>`Sec-CH-UA-Model`</dfn> request header field gives a server information about
-the device on which a given [=user agent=] is executing. It is a [=Structured Header=] whose value MUST
-be a [=structured header/string=] [[!RFC8941]].
+the device on which a given [=user agent=] is executing. It is a [=Structured Header=] whose value
+MUST be a [=structured header/string=] [[!RFC8941]].
 
 The header's ABNF is:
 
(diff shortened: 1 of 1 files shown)


miketaylr:
Mind taking a look @recvfrom?

recvfrom:
@miketaylr this looks good to me with a few comments (feel free to take them or leave them!):

 - My initial thought was that "querying the `Windows.Foundation.UniversalApiContract` integer version" was something only UWP apps could do easily, since API contracts seem like a UWP concept.  However, from looking at the Chrome source [1] it seems that this can be achieved by reading the value from the `SOFTWARE\Microsoft\WindowsRuntime\WellKnownContracts\Windows.Foundation.UniversalApiContract` registry key.  This approach seems undocumented AFAICT, so it seems like it could be challenging for implementers reading this spec to determine how to accomplish this step. Would it make sense for the spec to mention how to do this the way it does for the "get the legacy Windows version number" step? Or maybe just add a NOTE pointing to the Chrome source?

 - In the "get the legacy Windows version number" section, it might be helpful to annotate what versions of Windows the version strings "0.1", "0.2", and "0.3" correspond to, or link to [2] or something.

[1] https://source.chromium.org/chromium/chromium/src/+/main:components/embedder_support/user_agent_utils.cc;l=91;drc=b235f34b2f135c8af5e619157ae11161d539d850
[2] https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfoa#remarks

miketaylr:
Thanks for the feedback!
> Would it make sense for the spec to mention how to do this the way it does for the "get the legacy Windows version number" step? Or maybe just add a NOTE pointing to the Chrome source?

I feel like this won't be super hard to find by other implementers, and given that Chromium might change exact implementation details in the future, I think I'd prefer to leave out this particular registry key or a link to Chromium's source. 


> * In the "get the legacy Windows version number" section, it might be helpful to annotate what versions of Windows the version strings "0.1", "0.2", and "0.3" correspond to, or link to [2] or something.

Good idea!

PR #295, opened 2022-04-01
Fixes #292 - Add additional Fingerprinting guidance for GREASE

### LINKED ISSUE #292: GREASE-like UA brand lists should be careful not to introduce new fingerprinting surfaces
The [explainer](https://github.com/WICG/ua-client-hints#should-the-ua-string-really-be-a-set) and [spec](https://wicg.github.io/ua-client-hints/#grease) mention that a GREASE-like approach can be used to prevent abuse of brand lists. 

While this should help prevent such abuse, care should be taken so that the GREASE-like UA brand lists don't themselves introduce a new fingerprinting vector. For instance, if each client computes a random GREASE-like UA brand at startup that will be used for all navigations in all tabs, this could serve as such a fingerprinting vector. However, if the identifier is consistent for many users (all users of browser version X have the same GREASE-like UA brand) or unstable (a new GREASE-like UA brand is generated every network request in an unpredictable manner), then the GREASE-like UA brand won't serve as an effective fingerprinting vector.

The explainer, spec, and privacy / security questionnaire should all mention this risk and require implementations mitigate this fingerprinting scenario.

miketaylr:
PTAL @yoavweiss

***
<a href="https://pr-preview.s3.amazonaws.com/miketaylr/ua-client-hints/pull/295.html" title="Last updated on Apr 1, 2022, 8:07 PM UTC (99b6d99)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/295/639034a...miketaylr:99b6d99.html" title="Last updated on Apr 1, 2022, 8:07 PM UTC (99b6d99)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -992,9 +992,8 @@ However, it will still be possible for some, or all, hints to be requested and u
 
 [=User agents=] should take care to not introduce fingerprinting vectors through GREASE-like brand
 lists that might be unique for an individual or a very small group of users. Rather, a strategy
-should be employed where the arbitrary brand is either shared across many users (e.g., stable across
-major versions for all users) or unstable and unpredictable (e.g., regenerated per-request). However
-see [[#grease]] for balancing the needs of reducing cache variance and analystics use-cases.
+should be employed where the arbitrary brand is shared across many users (e.g., stable across
+major versions for all users).
 
 Access Restrictions {#access}
 -------------------
--- a/index.bs
+++ b/index.bs
@@ -990,6 +990,12 @@ However, it will still be possible for some, or all, hints to be requested and u
 [=user agents=] should consider policies to restrict or reduce access to parties that are known to
 <a lt="active fingerprinting">actively fingerprint</a> their users.
 
+[=User agents=] should take care to not introduce fingerprinting vectors through GREASE-like brand
+lists that might be unique for an individual or a very small group of users. Rather, a strategy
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -990,6 +990,12 @@ However, it will still be possible for some, or all, hints to be requested and u
 [=user agents=] should consider policies to restrict or reduce access to parties that are known to
 <a lt="active fingerprinting">actively fingerprint</a> their users.
 
+[=User agents=] should take care to not introduce fingerprinting vectors through GREASE-like brand
+lists that might be unique for an individual or a very small group of users. Rather, a strategy
+should be employed where the arbitrary brand is either shared across many users (e.g., stable across
+major versions for all users) or unstable and unpredictable (e.g., regenerated per-request). However
```
yoavweiss, line 996: Do we need the "unstable and unpredictable" variant, given that it'd be caching hostile?

miketaylr: I think I'll remove that. In theory a UA could take a stance where randomization is more important than cache-hits, but maybe we shouldn't encourage it (especially since per-release randomization is good enough IMHO).

miketaylr:
OK, let's try again @yoavweiss.

yoavweiss (approved):
LGTM

miketaylr:
thanks!

Issue #148, opened 2020-11-02
Use cases are only described in the explainer


othermaciej:
The explainer [extensively discusses use cases](https://github.com/WICG/ua-client-hints#use-cases) but they are mostly not mentioned in the spec. It would be good for some of this information to be in the spec, as it may affect implementer choices, and the spec should stand alone without needing to reference the explainer.

PR #303, opened 2022-05-23
Fixes #302 - Link to use cases from explainer

### LINKED ISSUE #302: De-dupe use cases in explainer/README
We can just link to the draft spec, that way there's less stuff to keep in sync.

miketaylr:
In case we add new ones, or edit them - we don't need to worry about keeping them in sync.

PTAL @yoavweiss


yoavweiss:
What's the reasoning, given that this is how things started before https://github.com/WICG/ua-client-hints/issues/148?

yoavweiss:
OK, looked at #302 and it's true that keeping things in sync can be a burden.

yoavweiss (approved):
LGTM

PR #305, opened 2022-06-10
"typically exposed" is hard to determine in a technical specification…

jwrosewell:
… and needs to be removed.

Why shouldn't a user agent return information about the model associated with a device that does not also return true for the Mobile field? Google's own services display the model of Windows desktop or laptop device when providing information about login history for example.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 10, 2022, 10:24 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fjwrosewell%2Fua-client-hints%2F021bc1b2c87b509e00f2eb9fc3c55daea13ded8f%2Findex.bs&md-warning=not%20ready)

```
FATAL ERROR: Obsolete biblio ref: [rfc7231] is replaced by [rfc9110]. Either update the reference, or use [rfc7231 obsolete] if this is an intentionally-obsolete reference.
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ua-client-hints%23305.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -478,10 +478,6 @@ or [=platform bitness=] unless the user's platform is one where both the followi
  * Different CPU architectures are likely to require different binary executable resources, and
      different binary executable resources are likely to be available.
 
-[=User Agents=] MUST return the empty string for [=model=] if [=mobileness=] is false. [=User
-Agents=] MUST return the empty string for [=model=] even if [=mobileness=] is true, except on
-platforms where the model is typically exposed.
-
 [=User agents=] MAY return the empty string for hints of type `sf-string`,
 `false` for hints of type `sf-boolean`, or any other fictitious value, for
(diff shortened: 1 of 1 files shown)


miketaylr:
Hi @jwrosewell, please open an issue for discussion first (including the use cases you have in mind). PRs that delete normative requirements (i.e., anything with a MUSIC) without prior discussion are very unlikely to be productive.

miketaylr:
See also https://github.com/WICG/ua-client-hints/blob/main/CONTRIBUTING.md#pull-requests

PR #310, opened 2022-06-22
Fixes #309 - Don't start or end a greaseyBrand w/ a greaseyChar

### LINKED ISSUE #309: GREASE: Consider disallowing special characters at index 0
We've identified a few cases in which a leading special character causes issues with WAF or other software (in one case, 403s were observed on a major bank website when `/` was leading the GREASE string, but not when `/` was present at other indices; in another, a job hunting site rejected requests with `;` as the first character, but not in other indices). [The spec](https://wicg.github.io/ua-client-hints/#create-arbitrary-brands-section) currently calls for a character, then a brand component, followed by stripping leading and trailing whitespace, such that index zero will usually be a special character.

I am sympathetic to firewall authors who would think leading with `=` or `/` looks like an attempt at tampering, and based on finding web incompatibilities with leading special characters, it may not be worth the risk. Note that keeping the same set of special characters, but keeping them outside the leading index in the string, still prevents blanket rejection of those characters or ossification on specific strings, so it still appears to support the goal of GREASE.

miketaylr:
Implementation experience has taught us that starting a brand w/
certain characters causes breakage (likely from firewalls). To
be conservative, we prevent starting or ending with those.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 28, 2022, 8:54 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FWICG%2Fua-client-hints%2Ffb52ee6a9022d346d6a45d4fc6ab92a926951076%2Findex.bs&md-warning=not%20ready)

```
FATAL ERROR: Obsolete biblio ref: [rfc7231] is replaced by [rfc9110]. Either update the reference, or use [rfc7231 obsolete] if this is an intentionally-obsolete reference.
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/ua-client-hints%23310.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -850,13 +850,15 @@ To <dfn>create an arbitrary brand</dfn>, the [=user agent=] MUST run these steps
     1. Let |greaseyStack| be a [=stack=].
     1. Let |greaseyChars| be the [=list=] of [=ASCII bytes=] « 0x20 (SP), 0x28 (left parenthesis), 0x29
          (right parenthesis), 0x2D (-), 0x2E (.), 0x2F (/), 0x3A (:), 0x3B (;), 0x3D (=), 0x3F (?), 0x5F (_) ».
-    1. For each item of |arbitraryBrandList|, [=stack/push=] a randomly selected [=list/item=] from |greaseyChars| onto
-         |greaseyStack|.
-    1. Let |greaseyBrandList| be a [=list=] and |index| be 0.
+    1. Let |index| be 0.
+    1. While |index| is less than the [=list/size=] of |arbitraryBrandList|:
+        1. [=stack/Push=] a randomly selected [=list/item=] from |greaseyChars| onto |greaseyStack|.
+        1. Increment |index| by 1.
+    1. Let |greaseyBrandList| be a [=list=] and set |index| to 0.
     1. While |greaseyStack| [=stack/is not empty=]:
-        1. Let |item| be the result of <a lt="pop">popping</a> from |greaseyStack|.
-        1. [=list/Append=] |item| to |greaseyBrandList|.
         1. [=list/Append=] |arbitraryBrandList|[|index|] to |greaseyBrandList|.
+        1. Let |item| be the result of [=popping=] from |greaseyStack|.
+        1. [=list/Append=] |item| to |greaseyBrandList|.
         1. Increment |index| by 1.
     1. Return the result of <a lt="strip leading and trailing ASCII whitespace"> stripping leading and trailing ASCII
          whitespace</a> from the [=concatenation=] of |greaseyBrandList| (with no separator).
--- a/index.bs
+++ b/index.bs
@@ -835,7 +835,7 @@ When asked to <dfn>create brands</dfn> with |version type|, run the following st
     Note: See [[#grease]] for more details on when and why these randomization steps might be appropriate.
 1. Return |list|.
 
-An <dfn for="user agent" export>equivalence class</dfn> represents a group of browsers believed to be compatibile with
+An <dfn for="user agent" export>equivalence class</dfn> represents a group of browsers believed to be compatible with
 each other. A shared rendering engine may form an [=equivalence class=], for example.
 
 <h4 id="create-arbitrary-brands-section"
@@ -867,7 +867,7 @@ To <dfn>create an arbitrary brand</dfn>, the [=user agent=] MUST run these steps
         separator).
 
     Note: This algorithm should result in an arbitrary brand without leading or trailing
-      |greaseyChars|, as implmentation experience has shown these are not web-compatibile.
+      |greaseyChars|, as implementation experience has shown these are not web-compatible.
 
 To <dfn>create an arbitrary version</dfn> given |version type|, run the following steps:
     1. [=Assert=] |version type| is either "full version" or "significant version".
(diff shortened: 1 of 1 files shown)


miketaylr:
Can you review @mreichhoff?

index.bs
```diff
@@ -850,13 +850,15 @@ To <dfn>create an arbitrary brand</dfn>, the [=user agent=] MUST run these steps
...
+    1. While |index| is less than the [=list/size=] of |arbitraryBrandList|:
+        1. [=stack/Push=] a randomly selected [=list/item=] from |greaseyChars| onto |greaseyStack|.
+        1. Increment |index| by 1.
+    1. Let |greaseyBrandList| be a [=list=] and set |index| to 0.
     1. While |greaseyStack| [=stack/is not empty=]:
-        1. Let |item| be the result of <a lt="pop">popping</a> from |greaseyStack|.
-        1. [=list/Append=] |item| to |greaseyBrandList|.
         1. [=list/Append=] |arbitraryBrandList|[|index|] to |greaseyBrandList|.
+        1. Let |item| be the result of [=popping=] from |greaseyStack|.
+        1. [=list/Append=] |item| to |greaseyBrandList|.
```
mreichhoff, line 861: wouldn't this still end up with a trailing `greaseyChar`? Unless the prior while loop's changes are meant to be equivalent to `for(i=0; i < arbitraryBrandList.size-1; i++)`, in which case I think we'd end up leaving off the last item from `arbitraryBrandList`?

miketaylr: Ah yeah, thanks. PTAL at the updated commits.

miketaylr:
@mreichhoff ready for round 2

index.bs
```diff
@@ -843,26 +843,31 @@ each other. A shared rendering engine may form an [=equivalence class=], for exa
 
 To <dfn>create an arbitrary brand</dfn>, the [=user agent=] MUST run these steps:
 
-    1. Let |arbitraryBrand| be a [=/string=] composed of [=ASCII alpha=]. |arbitraryBrand| MUST contain one or more
-         0x20 (SP) bytes and be no longer than twenty [=ASCII bytes=].
-    1. Let |arbitraryBrandList| be the result of <a lt="split on ASCII whitespace">splitting |arbitraryBrand| on ASCII
-         whitespace</a>.
+    1. Let |arbitraryBrand| be a [=/string=] composed of [=ASCII alpha=]. |arbitraryBrand| MUST
+        contain one or more 0x20 (SP) bytes and be no longer than twenty [=ASCII bytes=].
```
mreichhoff, line 847: what if arbitraryBrand had a space first? So then split would return `["","whatever"]`? I think you'd end up with a leading special character then. This might be fine though, given it says "composed of ASCII alpha", and I guess we could also leave open the possibility someone could very explicitly decide to start their string with a special character if they wanted to deal with it.

mreichhoff, line 847: what if arbitraryBrand had a space first? So then split would return `["","whatever"]`? I think you'd end up with a leading special character then. This might be fine though, given it says "composed of ASCII alpha", and I guess we could also leave open the possibility someone could very explicitly decide to start their string with a special character if they wanted to deal with it.

index.bs
```diff
@@ -843,26 +843,31 @@ each other. A shared rendering engine may form an [=equivalence class=], for exa
...
+        1. [=list/Append=] |item| to |greaseyBrandList|.
         1. Increment |index| by 1.
-    1. Return the result of <a lt="strip leading and trailing ASCII whitespace"> stripping leading and trailing ASCII
-         whitespace</a> from the [=concatenation=] of |greaseyBrandList| (with no separator).
+    1. [=list/Append=] |arbitraryBrandList|[|index|] to |greaseyBrandList|.
+    1. Return the result of <a lt="strip leading and trailing ASCII whitespace"> stripping leading
+        and trailing ASCII whitespace</a> from the [=concatenation=] of |greaseyBrandList| (with no
+        separator).
 
-    Note: [=Structured Headers=] allows for escaped 0x22 (\") and 0x5C (\\) inside a [=structured header/string=], but
```
mreichhoff, line 864: seems like we could keep this comment in addition to the next one?

index.bs
```diff
@@ -843,26 +843,31 @@ each other. A shared rendering engine may form an [=equivalence class=], for exa
...
-          these are known to not be web-compatible.
+    Note: This algorithm should result in an arbitrary brand without leading or trailing
+      |greaseyChars|, as implmentation experience has shown these are not web-compatibile.
```
mreichhoff, line 870: mega nit: compatible is misspelled

mreichhoff (commented):
I think this is fine, just a couple of minor comments now. Not sure if we LGTM % nits or how you'd like to run it...

index.bs
miketaylr: that's an amazing typo.

miketaylr: (jeeze, i misspelled "implementation" too.)

index.bs
miketaylr: That is a very good point. Let me update - let's make sure we don't start or end the string with 0x20.

mreichhoff (approved):
LGTM

miketaylr:
Thanks @mreichhoff!

PR #373, opened 2024-12-20
Fixes #151 - Define a "ch-ua-high-entropy-values" permission policy

### LINKED ISSUE #141: Examples for "platform architecture" don't match the actually allowed values

### LINKED ISSUE #151: getHighEntropyValues should not return all values, only values that have been opted in

miketaylr:
The allows top-level sites to restrict which third-party sites are able to receive high-entropy ua client hint data, or even disable itself from collection of any high-entropy ua client hint data.

***
<a href="https://pr-preview.s3.amazonaws.com/miketaylr/ua-client-hints/pull/373.html" title="Last updated on Dec 20, 2024, 2:43 AM UTC (2760d9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/ua-client-hints/373/7acf88f...miketaylr:2760d9a.html" title="Last updated on Dec 20, 2024, 2:43 AM UTC (2760d9a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1031,7 +1031,7 @@ This specification defines a [=policy-controlled feature=] identified by the str
 via the {{NavigatorUAData/getHighEntropyValues(hints)/hints|getHighEntropyValues()}} API.
 
 Note: If a given document is not allowed to use the `"ch-ua-high-entropy-values"` feature, the
-`getHighEntropyValues()` API will continue to return the low-entropy values.
+`getHighEntropyValues()` API will continue to return the low-entropy values for convenience.
 
 Security and Privacy Considerations {#security-privacy}
 ===================================
--- a/index.bs
+++ b/index.bs
@@ -338,7 +338,7 @@ brand of the browser to the user. These messages often also include the platform
 order to make sure the user knows which device is in question.
 
 Since such messaging doesn't require any server-side adaptation, it's better for this case to use
-the `userAgentData.getHighEntropyData()` method in order to retrieve the required information.
+the `userAgentData.getHighEntropyValues()` method in order to retrieve the required information.
 
 ### Download of appropriate binary executables ### {#binary-executable-downloads}
 Some sites are used to download binary executables of native applications, and need to be able to
@@ -373,7 +373,7 @@ the lower entropy values available through `Sec-CH-UA` for logging purposes, or 
 higher-entropy hints. The latter doesn't seem like something services should do just for forensic
 purposes. On the other hand, when specific issues are encountered, it may make sense for those
 services to opt-in to receive more details on the user agent, or use the
-`userAgentData.getHighEntropyData()` API for that purpose.
+`userAgentData.getHighEntropyValues()` API for that purpose.
 
 ### Fingerprinting ### {#fingerprinting-use-case}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -971,21 +971,25 @@ On getting, the {{NavigatorUAData/platform}} attribute must return the [=user ag
 
 <h4 id="getHighEntropyValues"><code>getHighEntropyValues</code> method</h4>
 
-The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></dfn> method MUST run these steps:
+The <dfn method for="NavigatorUA"><code>getHighEntropyValues(|hints|)</code></dfn> method MUST run
+these steps:
 
 1. Let |p| be a [=a new promise=] created in the [=current realm=].
 
-2. If the [=user agent=] decides one or more values in |hints| should not be returned, then [=reject=] and return |p| with a "{{NotAllowedError}}".
```
miketaylr, line 978: This has never been implemented, FWIW (at least not in Chromium).

index.bs
```diff
@@ -338,7 +338,7 @@ brand of the browser to the user. These messages often also include the platform
 order to make sure the user knows which device is in question.
 
 Since such messaging doesn't require any server-side adaptation, it's better for this case to use
-the `userAgentData.getHighEntropyData()` method in order to retrieve the required information.
+the `userAgentData.getHighEntropyValues()` method in order to retrieve the required information.
```
yoavweiss, line 341: Is this rename web compatible?

miketaylr: Good question. I was also a bit confused when I saw that :)

Turns out we just had 2 instances of _the wrong name_ in the use-cases section (🙈). Not sure when that crept it, or if getHighEntropyData was the original name way back when. But we're not actually changing any IDL - `getHighEntropyValues` is the thing we shipped (and continue to ship).

yoavweiss (approved):
LGTM

miketaylr:
thanks!

